PR #1, opened 2019-09-05
mkinitcpio: Produce reproducible initramfs images

esotericnonsense:
We achieve this by stripping timestamps from within the filesystem,
and by using a pipeline to strip inodes from the cpio archive.

It functions for at least the 'gzip', 'xz', 'bzip2', 'lz4' and 'cat'
compressors. The 'lzop' compressor embeds a runtime timestamp.

Motivation: https://reproducible-builds.org

Signed-off-by: Daniel Edgecumbe <git@esotericnonsense.com>

### DIFF
--- a/man/mkinitcpio.8.txt
+++ b/man/mkinitcpio.8.txt
@@ -301,6 +301,22 @@ These are only the variables that the core of mkinitcpio honor. Additional
 hooks may look for other environment variables and should be documented by the
 help output for the hook.
 
+Reproducibility
+---------------
+mkinitcpio aims to create reproducible initramfs images by default.
+
+This means that two subsequent runs of mkinitcpio should produce two files
+that are identical at the binary level.
+
+Timestamps within the initramfs are set to the Unix epoch of 1970-01-01.
+
+Note that in order for the build to be fully reproducible, the compressor
+specified (e.g. gzip, xz) must also produce reproducible archives. At the time
+of writing, as an inexhaustive example, the lzop compressor is incapable of
+producing reproducible archives due to the insertion of a runtime timestamp.
+
+More information can be found at https://reproducible-builds.org.
+
 Files
 -----
 '/etc/mkinitcpio.conf'::
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -211,25 +211,28 @@ build_image() {
             ;;
     esac
 
-    cpio_opts=('-0' '-o' '-H' 'newc')
-    (( _optquiet )) && cpio_opts+=('--quiet')
-    if (( EUID != 0 )); then
-        warning 'Not building as root, ownership cannot be preserved'
-        cpio_opts+=('-R' '0:0')
-    fi
-
     pushd "$BUILDROOT" >/dev/null
-    find . -mindepth 1 -printf '%P\0' |
-            LANG=C bsdcpio "${cpio_opts[@]}" |
+
+    # Reproducibility: set all timestamps to 0
+    find . -mindepth 1 -execdir touch -hcd "@0" "{}" +
+
+    find . -mindepth 1 -printf '%P\0' | sort -z |
+            LANG=C bsdtar --null -cnf - -T - |
+            LANG=C bsdtar --uid 0 --gid 0 --null -cf - --format=newc @- |
             $compress "${COMPRESSION_OPTIONS[@]}" > "$out"
+
     pipesave=("${PIPESTATUS[@]}") # save immediately
     popd >/dev/null
 
     if (( pipesave[0] )); then
         errmsg="find reported an error"
     elif (( pipesave[1] )); then
-        errmsg="bsdcpio reported an error"
+        errmsg="sort reported an error"
     elif (( pipesave[2] )); then
+        errmsg="bsdtar (step 1) reported an error"
+    elif (( pipesave[3] )); then
+        errmsg="bsdtar (step 2) reported an error"
+    elif (( pipesave[4] )); then
         errmsg="$compress reported an error"
     fi
 


esotericnonsense:
It may make sense to hold off on this for now as I think a SOURCE_DATE_EPOCH based solution[1] may be better as it could allow us to instantiate pacman with the variable set and thus automatically use reproducible builds where necessary.

I'm in the process of adjusting the archiso releng scripts and jamming the --reproducible flag in everywhere is quite clunky.

[1] https://reproducible-builds.org/docs/source-date-epoch/

esotericnonsense:
OK, @grazzolini this should be good for review now.

The change here means that `linux.preset` or all invocations of mkinitcpio don't need to be hacked as long as the env variable is set at some point (e.g. when invoking archiso build scripts).

It's slightly cheeky to use SOURCE_DATE_EPOCH to toggle the cpio --reproducible flag.

However, I can't see any reason not to use cpio --reproducible in the general case, other than the additional dependency. I don't think anyone will be negatively affects or surprised by this change.

falconindy:
FWIW, I originally chose bsdcpio over cpio because everyone on Arch already has libarchive installed. cpio is an added dependency that will need to be bumped up from `[extra]` to `[core]`.

Would be great if libarchive could add support for this sort of thing, but I don't expect any progress on that front (https://github.com/libarchive/libarchive/issues/975 for one of the problems).

esotericnonsense:
For the most part this was just a lazy fix. I can investigate the suggested solution there actually - sounds like using an interim step to filter out the inodes could work.

grazzolini:
@esotericnonsense, I think the code itself is nice, but I think adding cpio as a dependency will probably never fly, as @falconindy mentioned. To be honest, that was the only red flag I saw. I didn't test it yet, but I'll test this over this weekend.

esotericnonsense:
I've had some limited success using a double bsdtar step (as stated in the issue @falconindy linked) to strip the inode numbers. It produces an initramfs that boots fine etc.

For some reason though, at the moment it ends up producing a 90MiB cpio that decompresses to 27MiB. Tinkering continues.

esotericnonsense:
Right, OK, I've cracked it. Needed a --norecurse flag on bsdtar to prevent it from adding directories multiple times (because the find step would specify them multiple times).

mkinitcpio
```diff
@@ -217,26 +217,21 @@ build_image() {
...
-    cpio_opts=('-0' '-o' '-H' 'newc')
-    (( _optquiet )) && cpio_opts+=('--quiet')
-    if (( SOURCE_DATE_EPOCH )); then
-        cpio_bin="cpio"
-        cpio_opts+=('--reproducible')
-    else
-        cpio_bin="bsdcpio"
-    fi
+    cpio_opts=('--null' '-cf-' '--format=newc')
+    # TODO (( _optquiet )) && cpio_opts+=('--quiet')
```
esotericnonsense: RFC. On my box, bsdtar is silent anyway.
I suppose we want this to appent -v if not present?

esotericnonsense:
Need to adjust the PIPESTATUS/pipesave bits as the pipeline has more steps now

esotericnonsense:
PIPESTATUS and verbosity fixed, this should be good to go now with just libarchive required.

falconindy:
1) you've still got a check for requiring cpio.
2) I see no reason to make this non-default behavior. Two possibilities. The easy route of assuming `SOURCE_DATE_EPOCH=0`, or the slightly less easy route of assuming `SOURCE_DATE_EPOCH=$(time_of_mkinitcpio_build)`
3) Please don't warn on using lzop. Instead, this feature should be documented in the manpage in a new section on reproducibility. mkinitcpio's support for compression is fully arbitrary and the user could pass something that isn't quite lzop, but eventually uses lzop.

This all should be squashed into a single commit.

esotericnonsense:
1) ack, amusingly I caught this just before you commented. :)

2) SOURCE_DATE_EPOCH=$(time_of_mkinitcpio_build) is the way to go.

3) ack.

I'll work on those and squash.

esotericnonsense:
Cleaned up and squashed. Your comments should be addressed now.

Turns out gzip doesn't require `-n` when reading from stdin, so we can skip that as well.

It's actually a lot cleaner assuming we can default this behaviour.

I'm testing the functionality at the moment using a full archiso build.

man/mkinitcpio.8.txt
```diff
@@ -100,6 +100,19 @@ Options
 *-z, \--compress* 'compress'::
 	Override the compression method with the 'compress' program.
 
+Environment
+-----------
+The following environment variables influence the program behavior:
```
falconindy: But the whole point was to make this enabled by default, without the need to set SOURCE_DATE_EPOCH.

mkinitcpio
```diff
@@ -211,25 +211,33 @@ build_image() {
...
-        cpio_opts+=('-R' '0:0')
+        cpio_opts+=('--uid' '0' '--gid' '0')
     fi
 
     pushd "$BUILDROOT" >/dev/null
-    find . -mindepth 1 -printf '%P\0' |
-            LANG=C bsdcpio "${cpio_opts[@]}" |
+
+    # Reproducibility: set timestamps if SOURCE_DATE_EPOCH present
+    find . -mindepth 1 -printf '%P\0' | xargs -0 touch -hcd "@${SOURCE_DATE_EPOCH:-$(date +%s)}"
```
falconindy: No need for xargs here, just use `find ... -exec touch -hcd "$SOURCE_DATE_EPOCH" {} +`

mkinitcpio
```diff
@@ -211,25 +211,33 @@ build_image() {
             ;;
     esac
 
-    cpio_opts=('-0' '-o' '-H' 'newc')
-    (( _optquiet )) && cpio_opts+=('--quiet')
+    cpio_opts=('--null' '-cf-' '--format=newc')
+    (( _optquiet )) || cpio_opts+=('-v')
```
falconindy: Let's not do this -- just drop the opportunity to add -v entirely. At best, it'll just be nonsense on stderr. at worst, i'm not even sure... I would seriously doubt that it'll write to stdout, but this doesn't seem like nice default behavior.

mkinitcpio
```diff
@@ -211,25 +211,33 @@ build_image() {
...
     if (( EUID != 0 )); then
         warning 'Not building as root, ownership cannot be preserved'
-        cpio_opts+=('-R' '0:0')
+        cpio_opts+=('--uid' '0' '--gid' '0')
```
falconindy: We should probably unconditionally do this if we're committing to reproducible builds. There's little value in permissions in early userspace. There might not even be an /etc/passwd or /etc/group to map numeric IDs back to names.

esotericnonsense: Agreed.

mkinitcpio
esotericnonsense: Sure.

mkinitcpio
esotericnonsense: ack

man/mkinitcpio.8.txt
esotericnonsense: As per your earlier comment:

> The easy route of assuming SOURCE_DATE_EPOCH=0, or the slightly less easy route of assuming SOURCE_DATE_EPOCH=$(time_of_mkinitcpio_build)

The former can produce reproducible builds by default. If we go with that approach, we don't really need SOURCE_DATE_EPOCH at all. If no-one cares about timestamps in initramfs (I don't know of a reason to care about them) then we can just do that.

If we want the default to be the time of the mkinitcpio build (latest commit has this behaviour), then the default behaviour is not a reproducible build by definition right, if you run it twice you'll have different embedded timestamps.

esotericnonsense: To distill this: is there any reason to care about timestamps at all here, if not, let's just touch -hcd @1 or whatever, strip all references to S_D_E and be done with it?

falconindy: That's what I was trying to get at, but I wrote this hastily and my intent wasn't clear. I suggested something like the build date of mkinitcpio rather than 0 for two reasons:

1) lack of user surprise. This is wishy washy, but some people might be oddly concerned about timestamps from 1970.
2) encoding the build timestamp might give a hint as to what version of mkinitcpio created the archive. However, that's rather redundant -- we already have a VERSION file in the cpio.

I guess I really don't care what the value is -- I just don't see the point of making this some sort of opt-in feature. If we can make the initramfs reproducible in the common case without side effects, I see no reason not to.

So yes, IMO, strip everything related to SOURCE_DATE_EPOCH and document the fact that mkinitcpio tries to make reproducible archives by default. You can call out compression as something that might make the archive *not* reproducible and mention "known good recipes".

esotericnonsense: Gotcha. I hadn't realised you meant the build date of mkinitcpio itself rather than the runtime date.

eli-schwartz:
On the topic of "default reproducible or not", note that makepkg sets up the variable SOURCE_DATE_EPOCH in early runtime, if it does not already exist (using `$(date +%s)`), then simply makes use of that variable when doing internal accounting or `touch`ing files for bsdtar -c to consume.

In the case of makepkg, we want to push the variable into the environment for build systems to use and unify on, so I don't know if that's directly applicable here. Is there anything else which needs the variable set for reproducibility in mkinitcpio (other than that one bsdcpio invocation)?

(The idea is that if you run makepkg without SOURCE_DATE_EPOCH set, you can consult the embedded builddate and reproduce the package by, now, setting SOURCE_DATE_EPOCH.)

esotericnonsense:
@eli-schwartz nothing actually needs the SOURCE_DATE_EPOCH variable itself here. The bsdcpio invocation itself doesn't need it - mkinitcpio itself is modifying the timestamps based on its' presence.

libarchive doesn't mention it at all (bsdtar, bsdcpio,...)
GNU gzip has one mention of S_D_E, in `dfltcc.c` which is used to disable hardware compression on a specific platform and not relevant for us

esotericnonsense:
Right.

The latest commit sets all timestamps to 0 / 1970-01-01 and includes a comment on reproducibility in the manpage.

All references to SOURCE_DATE_EPOCH have been stripped.

I don't think it's necessary to futz around with the build date of mkinitcpio (as @falconindy states, the VERSION file is there anyway). Similarly, unless someone explicitly expresses a desire to change the timestamps within the file I don't think we need to support that as an option.

eli-schwartz:
Okay, great, if we don't actually need any time-based modifications other than to totally suppress file timestamps in bsdcpio then that makes things a lot simpler. If they don't serve a purpose then they... don't serve a purpose. :D

PR #2, opened 2019-09-10
Updated copyright year

Foxboron:
Arguably isn't important. But gotta have those sweet years.


jelly:
I don't think it's required since it's explicit but LGTM.

Foxboron:
Fixed :)

PR #3, opened 2019-09-10
Add flag to specify where to look for hooks.

Foxboron:
I'm aware that `mkinitcpio` does use `$PWD/{install,hook}`,  but it feels a bit silly that we are still looking through a path if we only want to be looking through a specific location.

The motivation for this is to look into creating a test suite for `mkinitcpio`. Specifying a hook directory for any additional hooks feels apt.

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -56,6 +56,7 @@ usage: ${0##*/} [options]
    -s, --save                   Save build directory. (default: no)
    -d, --generatedir <dir>      Write generated image into <dir>
    -t, --builddir <dir>         Use DIR as the temporary build directory
+   -D, --hookdir <dir>          Specify where to look for hooks.
    -V, --version                Display version information and exit
    -v, --verbose                Verbose output (default: no)
    -z, --compress <program>     Use an alternate compressor on the image
@@ -404,6 +405,11 @@ while :; do
         -r|--moduleroot)
             shift
             _optmoduleroot=$1
+            ;;
+        -D|--hookdir)
+            shift
+            _d_hooks="$(readlink -f "$1")/hooks"
+            _d_install="$(readlink -f "$1")/install"
             ;;
         --)
             shift
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -14,8 +14,7 @@ _f_functions=functions
 _f_config=mkinitcpio.conf
 _d_hooks="$PWD/hooks:/usr/lib/initcpio/hooks:/lib/initcpio/hooks"
 _d_install="$PWD/install:/usr/lib/initcpio/install:/lib/initcpio/install"
-_d_flag_hooks=""
-_d_flag_install=""
+_d_flag_{hooks,install}=
 _d_firmware=({/usr,}/lib/firmware/updates {/usr,}/lib/firmware)
 _d_presets=mkinitcpio.d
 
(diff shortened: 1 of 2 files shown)


mkinitcpio
```diff
@@ -405,6 +406,11 @@ while :; do
             shift
             _optmoduleroot=$1
             ;;
+        -D|--hookdir)
+            shift
+            _d_hooks="$(readlink -f "$1")/hooks"
+            _d_install="$(readlink -f "$1")/install"
```
falconindy: Why use readlink on these? Also, why not allow for multiple dirs?

Foxboron: Always felt proper to resolve the complete paths when taking user input.

And my use case doesn't involve multiple dirs, I think, so went the easy route.

falconindy: If you aren't going to do error checking on readlink, I don't particularly see the point. Dropping it also makes it simpler to support multiple paths.

grazzolini: It is user input anyway. I agree with @falconindy, unless we error check this, I don't see the need for readlink.

Foxboron: I had an attempt doing multiple hook paths, dropping `readlink` in the process.

mkinitcpio
```diff
@@ -408,8 +410,8 @@ while :; do
             ;;
         -D|--hookdir)
             shift
-            _d_hooks="$(readlink -f "$1")/hooks"
-            _d_install="$(readlink -f "$1")/install"
+            _d_flag_hooks+="$1/hooks:"
+            _d_flag_install+="$1/install:"
```
falconindy: You either need to build this as an array, or trim the trailing ":" when you reassign to _d_hooks, et al at the end. An empty element in the PATH refers to $PWD, which isn't something we want.

mkinitcpio
```diff
@@ -14,6 +14,8 @@ _f_functions=functions
 _f_config=mkinitcpio.conf
 _d_hooks="$PWD/hooks:/usr/lib/initcpio/hooks:/lib/initcpio/hooks"
 _d_install="$PWD/install:/usr/lib/initcpio/install:/lib/initcpio/install"
+_d_flag_hooks=""
+_d_flag_install=""
```
falconindy: No need to assign the empty string here, just `_d_flag_{install,hooks}=` is fine

mkinitcpio
```diff
@@ -423,6 +425,11 @@ if [[ -t 1 ]] && (( _optcolor )); then
     try_enable_color
 fi
 
+if [[ -n $_d_flag_hooks && -n $_d_flag_install ]]; then
```
falconindy: Shouldn't this be either or? Why not allow mix and match of hooks and install scripts?

mkinitcpio
Foxboron: Fixed

mkinitcpio
Foxboron: Fix'd

mkinitcpio
Foxboron: Needs a second flag and split up the logic. Don't see a usecase currently

Foxboron:
:shipit:

mkinitcpio
grazzolini: Yes, to split this, we would need a install hooks dir flag and a hooks dir flag. As I've mentioned before, since they are tightly coupled, there's no need for separating this. Unless we start allowing run hooks that do not have a install counterpart, I don't see a reason for splitting. Also, I don't see this flag being very useful outside of testing.

grazzolini:
@Foxboron, can you add the hooks dir option to the parser so we can merge this PR?

PR #7, opened 2019-10-09
Pacman hook and script

grazzolini:
Create a pacman hook and a script for replacing the ones that are bundled with each kernel.


libalpm/scripts/mkinitcpio-install
```diff
@@ -0,0 +1,19 @@
...
+              /etc/mkinitcpio.d/"${pkgbase}".preset
+        fi
+        install -Dm644 $(dirname $line)/vmlinuz /boot/vmlinuz-"${pkgbase}"
+    else
+        args=('-P') # all presets
+        break
+    fi
+done
+
+mkinitcpio "${args[@]}" "${pkgbase}"
```
heftig: This will run mkinitcpio for at most one `pkgbase`. I think what you want to start with an empty `args=()` and add `args+=(-p "$pkgbase")` each loop.

Also, I think the `if` check and the `else` branch doing `-P` is pretty pointless as you will always have a list `pkgbase` files as input.

heftig:
```bash
#!/bin/bash -e

args=()

while read -r line; do
    pkgbase=$(<"$line")
    preset="/etc/mkinitcpio.d/${pkgbase}.preset"
    if [[ ! -e $preset ]]; then
        sed "s|%PKGBASE%|${pkgbase}|g" /usr/share/mkinitcpio/hook.preset \
            | install -Dm644 /dev/stdin "$preset"
    fi
    install -Dm644 "${line%/pkgbase}/vmlinuz" "/boot/vmlinuz-${pkgbase}"
done

mkinitcpio "${args[@]}"
```

grazzolini: See the hook, it triggers on usr/lib/modules/*/pkgbase as well as on Package = mkinitcpio*. So, -P is there to trigger when mkinitcpio itself is being upgraded. Using mkinitcpio* also has the benefit of regenerating all presets when mkinitcpio-busybox or mkinitcpio-nfs-utils are upgraded as well. Not to mention my own packages that are on community.

heftig: Ah, right, the hook also matches on `mkinitcpio-*` packages... Still, we need to make sure to process all kernel targets.

In addition, I think it looks more correct for the hook to match on `vmlinuz` instead of `pkgbase`.

```bash
#!/bin/bash -e

args=()
all=0

while read -r line; do
    if [[ $line != */vmlinuz ]]; then
        all=1
        continue
    fi

    pkgbase=$(<"${line%/vmlinuz}/pkgbase")
    [[ -n ${pkgbase} ]]

    preset="/etc/mkinitcpio.d/${pkgbase}.preset"
    if [[ ! -e $preset ]]; then
        sed "s|%PKGBASE%|${pkgbase}|g" /usr/share/mkinitcpio/hook.preset \
            | install -Dm644 /dev/stdin "$preset"
    fi

    install -Dm644 "${line}" "/boot/vmlinuz-${pkgbase}"
    args+=(-p "${pkgbase}")
done

if (( all )); then
    args=(-P)
fi

mkinitcpio "${args[@]}"
```

```ini
[Trigger]
Type = Package
Operation = Install
Operation = Upgrade
Target = mkinitcpio*

[Trigger]
Type = File
Operation = Install
Operation = Upgrade
Target = usr/lib/modules/*/vmlinuz

[Action]
Description = Updating boot kernels and initcpios...
When = PostTransaction
Exec = /usr/share/libalpm/scripts/mkinitcpio-install
NeedsTargets
```

grazzolini: I've also just tested this hook installing two kernels at the same time, and it's working just for the first of them. The mkinitcpio run itself must be inside the loop.

grazzolini: Implemented by dfa609a6ee5f75af1aa5f94d331aea60b6d076ca

libalpm/scripts/mkinitcpio-install
```diff
@@ -0,0 +1,33 @@
...
+
+    install -Dm644 "${line}" "/boot/vmlinuz-${pkgbase}"
+    args+=(-p "${pkgbase}")
+done
+
+if (( all )); then
+    args=(-P)
+fi
+
+if ! compgen -G /etc/mkinitcpio.d/*.preset > /dev/null; then
```
heftig: You should quote the glob here. Otherwise you're passing a single (existing) file to `-G` and other files as additional arguments, and I'm not sure what the effect of that is.

```bash
if ! compgen -G '/etc/mkinitcpio.d/*.preset' > /dev/null; then
```

libalpm/scripts/mkinitcpio-install
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
heftig: I think we really should add `-e` here so that errors abort the script.

grazzolini: Solved both in 59bdecbf138936f1eb61c9616d7270d92a5deac5

libalpm/scripts/mkinitcpio-install
```diff
@@ -0,0 +1,33 @@
...
+        continue
+    fi
+
+    pkgbase=$(<"${line%/vmlinuz}/pkgbase")
+    [[ -n ${pkgbase} ]]
+
+    preset="/etc/mkinitcpio.d/${pkgbase}.preset"
+    if [[ ! -e $preset ]]; then
+        sed "s|%PKGBASE%|${pkgbase}|g" /usr/share/mkinitcpio/hook.preset \
+            | install -Dm644 /dev/stdin "$preset"
```
eli-schwartz: If some nonconformant kernel package does not do what it is expected to do and fails to install a mkinitcpio preset, we can just skip it. Please don't go install configuration files behind my back that aren't tracked by pacman -- it would cause file conflicts when the kernel package gets updated to include the mkinitcpio preset due to my complaining to the maintainer. :p

If you absolutely must support nonconforming packages, install the hook to /tmp or /run, and call mkinitcpio with the full absolute filename.

libalpm/scripts/mkinitcpio-install
```diff
@@ -0,0 +1,33 @@
...
+args=()
+all=0
+
+while read -r line; do
+    if [[ $line != */vmlinuz ]]; then
+        all=1
+        continue
+    fi
+
+    pkgbase=$(<"${line%/vmlinuz}/pkgbase")
```
eli-schwartz: mapfile is still fine. `read` would also be fine. Either one is fewer subprocesses.

libalpm/hooks/90-mkinitcpio-install.hook
```diff
@@ -0,0 +1,17 @@
+[Trigger]
+Type = Package
+Operation = Install
+Operation = Upgrade
+Target = mkinitcpio*
```
eli-schwartz: This is wrong and fails to implement feature compatibility with the current hook, which triggers on `usr/lib/initcpio/*`.

The old hook would trigger on the mkinitcpio package, any packages which were named mkinitcpio-*, and any packages which were named perfectly ordinary innocent things that had nothing to do with mkinitcpio, but also installed initcpio files.

The new hook only supports packages which ship their initcpio files in a split package named mkinitcpio-*. Here's the list of packages affected, most of which would be broken:

```
$ pkgfile -vd /usr/lib/initcpio/
eschwartz/zfs-utils 0.8.2-2           	/usr/lib/initcpio/
core/btrfs-progs 5.2.2-1              	/usr/lib/initcpio/
core/cryptsetup 2.2.1-1               	/usr/lib/initcpio/
core/device-mapper 2.02.186-2         	/usr/lib/initcpio/
core/dmraid 1.0.0.rc16.3-11           	/usr/lib/initcpio/
core/lvm2 2.02.186-2                  	/usr/lib/initcpio/
core/mdadm 4.1-1                      	/usr/lib/initcpio/
core/mkinitcpio 25-2                  	/usr/lib/initcpio/
core/mkinitcpio-busybox 1.30.1-1      	/usr/lib/initcpio/
core/mkinitcpio-nfs-utils 0.3-6       	/usr/lib/initcpio/
core/systemd 243.51-1                 	/usr/lib/initcpio/
extra/archboot 2019.03-1              	/usr/lib/initcpio/
extra/archiso 42-1                    	/usr/lib/initcpio/
extra/ostree 2019.3-3                 	/usr/lib/initcpio/
community/mkinitcpio-dropbear 0.0.3-5 	/usr/lib/initcpio/
community/mkinitcpio-netconf 0.0.5-1  	/usr/lib/initcpio/
community/mkinitcpio-tinyssh 0.0.4-3  	/usr/lib/initcpio/
community/mkinitcpio-utils 0.0.3-4    	/usr/lib/initcpio/
community/pacredir 0.4.2-1            	/usr/lib/initcpio/
```

eli-schwartz:
Aside: if some odd kernel package was determined to do things weirdly, it might decide to use some weird name for the kernel too (as opposed to `/boot/vmlinuz-$pkgname`). If mkinitcpio decides to become responsible for creating presets from scratch, should it not also become responsible for checking that everything is actually there?

libalpm/scripts/mkinitcpio-install
heftig: This goes along with changes to the kernels which remove all files outside `/usr/lib/modules/$kernver/`.

grazzolini: Well, we discussed this at length, and all the official kernels at least, will not have anything outside /usr/lib/modules. It will be either mkinitcpio's or dracut's job to handle the actual kernel installation and initramfs generation. I'm still working on dracut hooks and we are discussing even more changes for that. I guess I can add a check, for example, that if a package has vmlinuz (most will), but doesn't have the pkgbase file, we bail and let that kernel do their own thing. Which, might as well include their own mkinitcpio hook and preset.

libalpm/scripts/mkinitcpio-install
grazzolini: Sure, I'll change the implementation to use either mapfile or read.

libalpm/hooks/90-mkinitcpio-install.hook
grazzolini: This hook takes care of rebuild the initramfs when any mkinitcpio* package is updated, so, all packages starting with mkinitcpio- on this list are not affected, and I've tested this. I don't like all these packages touch /usr/lib/initcpio, but I guess we have to live with that. It's easy enough to change the hook to trigger on usr/lib/initcpio/*.

grazzolini:
> Aside: if some odd kernel package was determined to do things weirdly, it might decide to use some weird name for the kernel too (as opposed to `/boot/vmlinuz-$pkgname`). If mkinitcpio decides to become responsible for creating presets from scratch, should it not also become responsible for checking that everything is actually there?

We are only going to care about official kernels here. But again, I can add the check for pkgbase and bail if it's not there.

libalpm/scripts/mkinitcpio-install
heftig: Use `read -r pkgbase < "${line%/vmlinuz}/pkgbase"`.

`mapfile` is reading a set of lines into an array.

libalpm/scripts/mkinitcpio-install
heftig: Doing it this way decouples our kernels from the way they are installed into `/boot/`. This will allow us to implement better ways in `dracut` or other packages.

libalpm/scripts/mkinitcpio-install
eli-schwartz: The first revision read a set of lines into an array, using "append to end of array" mode:

```
mapfile -O ${#targets[@]} -t targets < "/$line"
```

Yes, if you want a non-array variable (the script has changed a lot since the initial one I hacked up) then read is your friend.

libalpm/scripts/mkinitcpio-install
eli-schwartz: So we're making tons of sweeping changes all at once, including moving around the vmlinuz file in ways that mess up peoples' expectations and don't even fit the purpose of lib/modules anyway, in order to implement mysterious "better ways" that are probably the inferior `kernel-install` which isn't even needed for dracut?

Not a fan at all, but either way I think we should make incremental changes, focus on making the hooks better, and then re-envision how we should install (or fail to install) kernels for all of arch.

grazzolini: I've added the check for non pkgbase kernels with 9c1e0f3dd01151b79e9b557ea8510605616c1369.

libalpm/scripts/mkinitcpio-install
grazzolini: Implemented on 9c1e0f3dd01151b79e9b557ea8510605616c1369

libalpm/hooks/90-mkinitcpio-install.hook
grazzolini: I have implemented this with 0791321ed88bc480b193889c26cf4c356f1c78f7. The hook triggers on any change to /usr/lib/initcpio.

libalpm/scripts/mkinitcpio-remove
```diff
@@ -0,0 +1,39 @@
...
+    preset="/etc/mkinitcpio.d/${pkgbase}.preset"
+    initramfs="/boot/initramfs-${pkgbase}.img"
+    fallback_initramfs="/boot/initramfs-${pkgbase}-fallback.img"
+    if [[ -e $kernel ]]; then
+        # remove the installed kernel
+        rm $kernel
+    fi
+    if [[ -e $preset ]]; then
+        # remove the preset
+        rm $preset
```
eli-schwartz: So we're back to deleting userdata? When this file was a pacman-managed configuration file, uninstalling a kernel would result in this becoming a pacsave, so you could reinstall the kernel and restore your custom configuration. Now deleting a kernel causes your modifications in /etc to be magically nuked.

grazzolini: Yes, for now. I'm going to rework this script once I add the option for mkinitcpio to remove the images. I'm going to add a check if this file was changed, by comparing it with one generated from the preset, and if it's different, I'll move it to .pacsave.

grazzolini:
> Why did the Makefile installer for libalpm/scripts/mkinitcpio-remove get added in the commit _before_ the commit which actually checked the script into git?

No particular reason, with the exception I only added the final version to the repo.

libalpm/scripts/mkinitcpio-remove
heftig: I still think just leaving the file in place is much easier than trying to duplicate the backup algorithm,

eli-schwartz: @heftig Trying to take over pacman's job means you need to take over pacman's job all the way. Leaving the hook behind means the preset is broken, and thus, `mkinitcpio --allpresets` is broken

eli-schwartz: @grazzolini Another solution is kernels can just... install them themselves?

It's an inert file, it does no harm if mkinitcpio is not installed... I don't understand why core/linux is trying so hard to strip itself of usefulness and make literally everything someone else's responsibility right down to actually installing the vmlinuz file somewhere usable.

heftig: @eli-schwartz You're right; I forgot about `mkinitcpio -P`.

eli-schwartz: Honestly if the worry is about duplicating the preset template file we could still install it with mkinitcpio and have the kernel makedepend on mkinitcpio to use that.

grazzolini: > @grazzolini Another solution is kernels can just... install them themselves?
> 
> It's an inert _configuration_ file, it does no harm if mkinitcpio is not installed since nothing reads it... I don't understand why core/linux is trying so hard to strip itself of usefulness and make literally everything someone else's responsibility right down to actually installing the vmlinuz file somewhere usable.

So, will the kernels also install a dracut configuration for each of them? I agree with the decision of removing anything mkinitcpio specific from the kernels. It was the right move.

eli-schwartz: I lack knowledge of how dracut operates, so I don't know if there's a comparable option. But sure, why not? There are absolutely tons of packages which install optional components that are only used by another package that may not be installed.

For example:

Lots of packages install /usr/lib/initcpio/ files, whether or not mkinitcpio is installed. These files do nothing, except advertise functionality if mkinitcpio is installed or is later installed.


Some packages do the same for dracut:

```
$ pkgfile -vr /usr/lib/dracut| grep -v ^extra/dracut
eschwartz/zfs-utils 0.8.2-2	/usr/lib/dracut/modules.d/02zfsexpandknowledge/module-setup.sh
eschwartz/zfs-utils 0.8.2-2	/usr/lib/dracut/modules.d/90zfs/export-zfs.sh
eschwartz/zfs-utils 0.8.2-2	/usr/lib/dracut/modules.d/90zfs/module-setup.sh
eschwartz/zfs-utils 0.8.2-2	/usr/lib/dracut/modules.d/90zfs/mount-zfs.sh
eschwartz/zfs-utils 0.8.2-2	/usr/lib/dracut/modules.d/90zfs/parse-zfs.sh
eschwartz/zfs-utils 0.8.2-2	/usr/lib/dracut/modules.d/90zfs/zfs-env-bootfs.service
eschwartz/zfs-utils 0.8.2-2	/usr/lib/dracut/modules.d/90zfs/zfs-generator.sh
eschwartz/zfs-utils 0.8.2-2	/usr/lib/dracut/modules.d/90zfs/zfs-lib.sh
eschwartz/zfs-utils 0.8.2-2	/usr/lib/dracut/modules.d/90zfs/zfs-load-key.sh
eschwartz/zfs-utils 0.8.2-2	/usr/lib/dracut/modules.d/90zfs/zfs-needshutdown.sh
community/clevis 11-6      	/usr/lib/dracut/modules.d/60clevis/clevis-hook.sh
community/clevis 11-6      	/usr/lib/dracut/modules.d/60clevis/module-setup.sh
community/nvme-cli 1.9-1   	/usr/lib/dracut/dracut.conf.d/70-nvmf-autoconnect.conf
community/tpm2-totp 0.2.0-1	/usr/lib/dracut/modules.d/70tpm2-totp/README
community/tpm2-totp 0.2.0-1	/usr/lib/dracut/modules.d/70tpm2-totp/cleanup-tpm2-totp.sh
community/tpm2-totp 0.2.0-1	/usr/lib/dracut/modules.d/70tpm2-totp/module-setup.sh
community/tpm2-totp 0.2.0-1	/usr/lib/dracut/modules.d/70tpm2-totp/show-tpm2-totp
community/tpm2-totp 0.2.0-1	/usr/lib/dracut/modules.d/70tpm2-totp/show-tpm2-totp.sh
```

Note: the zfs-utils AUR package installs files to both /usr/lib/initcpio/ and /usr/lib/dracut/ at the same time.

It seems perfectly fair to do mkinitcpio presets too. If dracut has the ability to register preset configurations, then I absolutely encourage the kernels to package both mkinitcpio and dracut presets.

PR #13, opened 2019-12-06
colorize and casper-like live boot support added

paledega:


paledega:
We are using this feature in SulinOS.
https://github.com/SulinOS/mkinitcpio-donkey

paledega:
and look like this (with openrc)
photo_2019-12-03_15-08-15

falconindy:
Much of this can be done through hooks. Why should it be a part of init directly?

Lots of sloppy code here, typos, and lack of style consistency with existing code.

Also, Arch is generally wanting to get away from mkinitcpio and is moving towards dracut.

grazzolini:
This belongs on a hook. I don't see any reason adding this to init directly. Also, even if we were to add this to init, if shouldn't be the default as long quiet is not there, and should have instead, it's own flag to trigger.

Not to mention that I'm not even sure what this does to the logs, because of all the escape chars.

PR #15, opened 2020-01-20
fix running as non-root creates non-root files in the cpio

gdamjan:
https://bugs.archlinux.org/task/65006

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -221,7 +221,7 @@ build_image() {
     # If this pipeline changes, |pipeprogs| below needs to be updated as well.
     find . -mindepth 1 -printf '%P\0' |
             sort -z |
-            LANG=C bsdtar --null -cnf - -T - |
+            LANG=C bsdtar --uid 0 --gid 0 --null -cnf - -T - |
             LANG=C bsdtar --uid 0 --gid 0 --null -cf - --format=newc @- |
             $compress "${COMPRESSION_OPTIONS[@]}" > "$out"
 


mkinitcpio
```diff
@@ -221,7 +221,7 @@ build_image() {
     # If this pipeline changes, |pipeprogs| below needs to be updated as well.
     find . -mindepth 1 -printf '%P\0' |
             sort -z |
-            LANG=C bsdtar --null -cnf - -T - |
+            LANG=C bsdtar --uid 0 --gid 0 --null -cnf - -T - |
             LANG=C bsdtar --uid 0 --gid 0 --null -cf - --format=newc @- |
```
gdamjan: dreisner suggested that this second line probably doesn't need the --uid/--gid, since it obviously doesn't work. should I remove those?

Foxboron: Might want to let @esotericnonsense look over and ensure it doesn't break reproducability.

grazzolini: > dreisner suggested that this second line probably doesn't need the --uid/--gid, since it obviously doesn't work. should I remove those?

This second line looks weird to me, indeed.

grazzolini: > Might want to let @esotericnonsense look over and ensure it doesn't break reproducability.

I'm going to test it also. And is reproducibility :-)

gdamjan: any news?

gdamjan:
:1234: ?

mkinitcpio
gdamjan: btw, doesn't the above `sort -z` also need a `LANG=C` option (just in case)?

grazzolini: @gdamjan
There are some more things to fix on that pipe, see #16. I'm doing a revamp of this pipeline some time this week, so you might need to rebase your patch after that.

grazzolini:
#38 implements the same feature with some more changes. I'm closing this one and I'll evaluate the other to see if it doesn't break reproducibility.

PR #16, opened 2020-02-12
Make build process and resulting images capable of storing secrets.

bkus-goog:
When storing secrets (eg: crypttab keyfiles) in initramfs, ensure the secrets are protected during the build process, and in the resulting initramfs images.  This works best when initramfs is also stored on an encrypted or otherwise secured filesystem.  Access by root is not a problem since "dmsetup table --showkeys" will show the keys to root anyway.

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -223,10 +223,11 @@ build_image() {
             sort -z |
             LANG=C bsdtar --null -cnf - -T - |
             LANG=C bsdtar --uid 0 --gid 0 --null -cf - --format=newc @- |
-            $compress "${COMPRESSION_OPTIONS[@]}" > "$out"
+            $compress "${COMPRESSION_OPTIONS[@]}" |
+            (rm -f "$out" && umask 077 && cat > "$out")
 
     pipestatus=("${PIPESTATUS[@]}")
-    pipeprogs=('find' 'sort' 'bsdtar (step 1)' 'bsdtar (step 2)' "$compress")
+    pipeprogs=('find' 'sort' 'bsdtar (step 1)' 'bsdtar (step 2)' "$compress" 'save image')
 
     popd >/dev/null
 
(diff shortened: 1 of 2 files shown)


bkus-goog:
Have you had a chance to review this?  As it only affects permissions of the temporary build directory and the resulting initramfs file, it should not have any reproducibility problems.

grazzolini:
I had not a chance to test this yet. But if you could test the reproducibility, it would be great.

bkus-goog:
Is there a specific process or script for that?  I'm not entirely clear on what you'd be looking for to prove reproducibility.  Just same archive contents before + after this change?

grazzolini:
Run mkinitcpio, take a hash of the resulting images. Delete them (for ensuring you're not seeing old results), run mkinitcpio again and take the hash again, they should be the same. mkinitcpio currently lacks a test for this (or much tests in general), but it's pretty straightforward to verify this.

bkus-goog:
Finally ran the test.  The hashes of main and fallback images are identical before and after patch application.

bkus-goog:
Is there anything left to do here, or can we merge?

mkinitcpio
```diff
@@ -223,10 +223,11 @@ build_image() {
             sort -z |
             LANG=C bsdtar --null -cnf - -T - |
             LANG=C bsdtar --uid 0 --gid 0 --null -cf - --format=newc @- |
-            $compress "${COMPRESSION_OPTIONS[@]}" > "$out"
+            $compress "${COMPRESSION_OPTIONS[@]}" |
+            (rm -f "$out" && umask 077 && cat > "$out")
```
falconindy: This isn't explicitly a feature of your patch, but it feels like the right time to suggest that we generate the image in the temporary $_d_workdir and only rotate it into place after successful generation. A failure anywhere in this pipeline causes the user to be left without a initramfs image. Not great.

Generating this separately also means we can avoid this awkward command group as part of the pipeline.

grazzolini:
@bkus-goog I'm going over PR's for a new mkinitpcio release. Is this still necessary? Also, there are a few unresolved questions, can you address those?

grazzolini:
Closing this one for now.

PR #19, opened 2020-03-03
alpm-hooks should use Type = Path, not File

sirlucjan:
With reference to similar changes in other packages (systemd, kmod, dbus) it would be good to change here as well.

Signed-off-by: Piotr Gorski <lucjan.lucjanov@gmail.com>


sirlucjan:
I'A glitch crept into PR, so I closed it and created a new one.

PR #17, opened 2020-02-20
[functions] Ignore modinfo built-in modules

Foxboron:
modinfo have started reporting filenames as "(builtin)" on built-in
modules. Ignore these and continue with other modules.

Fixes: https://bugs.archlinux.org/task/65564
Introduced in: https://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git/commit/?id=e7e2cb61fa9f1db3429d91ef6accff549500d268

Signed-off-by: Morten Linderud <morten@linderud.pw>

PR #20, opened 2020-03-07
Fixes for FS#65725

falconindy:


functions
```diff
@@ -376,11 +376,17 @@ add_module() {
     while IFS=':= ' read -r -d '' field value; do
         case "$field" in
             filename)
-                found=1
-                module=${value##*/} module=${module%.ko*}
-                quiet "adding module: %s" "$module"
-                _modpaths["$value"]=1
-                _addedmodules["${module//-/_}"]=1
+                # Only add modules with filenames that look like paths (e.g.
+                # it might be repored as "(builtin)"). We'll defer actually
```
eli-schwartz:
```suggestion
                # it might be reported as "(builtin)"). We'll defer actually
```

Maryse47:
ping @grazzolini

Enelar:
@falconindy #17

Maryse47:
@Enelar the PR you linked isn't sufficient to fix issue this PR fixes.

ItsiAdam:
This fixes the issue for me.

tlahdekorpi:
This patch makes it so that all builtin modules that have alternate hardware accelerated alternatives like sha256, crc32c or aes are excluded from the image. Most filesystems depend on one of these modules and even if the alternate modules are loaded after bootup they never get used.

falconindy:
@tlahdekorpi Please provide concrete examples.

tlahdekorpi:
When sysroot is on btrfs and it gets mounted during boot with crc32c or sha256 checksumming the alternate modules never get used.
```sh
$ cat install/test
build() { add_module btrfs; }
$ ./mkinitcpio -k 5.6.13-300.fc32.x86_64 -c /dev/null -A test -g /tmp/no-patch > /dev/null
$ ./mkinitcpio -k 5.6.13-300.fc32.x86_64 -c /dev/null -A test -g /tmp/patch > /dev/null
$ diff -u0 <(bsdtar -tf /tmp/no-patch) <(bsdtar -tf /tmp/patch)
--- /dev/fd/63	2020-05-19 19:58:34.687271190 +0000
+++ /dev/fd/62	2020-05-19 19:58:34.688271181 +0000
@@ -28 +27,0 @@
-usr/lib/modules/5.6.13-300.fc32.x86_64/kernel/crc32c-intel.ko
@@ -31 +29,0 @@
-usr/lib/modules/5.6.13-300.fc32.x86_64/kernel/sha256-ssse3.ko
```

falconindy:
Sounds like you were broken by 30dbfc2770cb49cb3babfb750205b126d7d88791, not anything in this pull request.

falconindy:
For posterity, we determined in IRC that https://github.com/archlinux/mkinitcpio/commit/30dbfc2770cb49cb3babfb750205b126d7d88791 is actually responsible and should be reverted on top of this pull request.

The problem is that https://github.com/archlinux/mkinitcpio/commit/30dbfc2770cb49cb3babfb750205b126d7d88791 causes builtins to not be resolved as potential aliases and leads to hardware-dependent optimized modules not being added when the core alias (e.g. crc32c vs. crc32c-intel) is present as a builtin,

PR #22, opened 2020-05-20
Improve module resolution and fix FS#65725

tlahdekorpi:
Include all builtin module softdeps, firmware and aliases. Reduce time spent running modinfo and report all missing firmware.

Fixes FS#65725

### DIFF
--- a/functions
+++ b/functions
@@ -397,6 +397,10 @@ _add_modules() {
                 ;;
             filename)
                 quiet "adding module file: %s" "$value"
+
+                # kmod v27 reports all builtin module filenames as (builtin).
+                [[ "$value" == '(builtin)' ]] && continue
+
                 module=${value##*/} module=${module%.ko*}
                 _modpaths["$value"]=1
                 _addedmodules["${module//-/_}"]=1
--- a/functions
+++ b/functions
@@ -312,7 +312,7 @@ add_all_modules() {
     local mod mods
 
     mapfile -t mods < <(all_modules "$@")
-    map add_module "${mods[@]}"
+    add_module "${mods[@]}"
 
     return $(( !${#mods[*]} ))
 }
@@ -330,7 +330,7 @@ add_checked_modules() {
         mapfile -t mods < <(all_modules "$@")
     fi
 
-    map add_module "${mods[@]}"
+    add_module "${mods[@]}"
 
     return $(( !${#mods[*]} ))
 }
@@ -676,7 +676,7 @@ add_binary() {
 parse_config() {
     # parse key global variables set by the config file.
 
-    map add_module "${MODULES[@]}"
+    add_module "${MODULES[@]}"
     map add_binary "${BINARIES[@]}"
     map add_file "${FILES[@]}"
 
--- a/functions
+++ b/functions
@@ -354,6 +354,35 @@ add_firmware() {
     done
 }
 
+add_builtin_modinfo() {
+    local module softdeps field value
+
+    while IFS='=' read -r -d '' field value; do
+        case "$field" in
+            *.firmware)
+                _firmware+=("$value")
+                ;;
+            *.alias)
+                _addedmodules["${value//-/_}"]=3
+                ;;
+            *.softdep)
+                read -ra softdeps <<<"$value"
+                for module in "${softdeps[@]}"; do
+                    [[ $module == *: ]] && continue
+                    add_module "$module?"
+                done
+                ;;
+            *.*)
+                # Don't rely on anything other than a entry for the module
+                # existing since nothing is guaranteed.
+                # e.g. not all modules have a .file entry.
+                module="${field%.*}"
+                _addedmodules["${module//-/_}"]=2
+                ;;
+        esac
+    done
+}
+
 _add_modules() {
     local m module softdeps deps field value
 
--- a/functions
+++ b/functions
@@ -336,58 +336,51 @@ add_checked_modules() {
 }
 
 add_firmware() {
-    # add a firmware file to the image.
-    #   $1: firmware path fragment
+    local fw p
 
-    local fw fwpath r=1
+    [[ $KERNELVERSION == none ]] && return
 
+    # "/lib/firmware/updates/" UTS_RELEASE,
+    # "/lib/firmware/updates",
+    # "/lib/firmware/" UTS_RELEASE,
+    # "/lib/firmware"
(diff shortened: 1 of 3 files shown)


functions
```diff
@@ -336,58 +336,47 @@ add_checked_modules() {
...
-    if [[ $1 = *\? ]]; then
-        ign_errors=1
-        set -- "${1%?}"
-    fi
-
-    target=${1%.ko*} target=${target//-/_}
-
-    # skip expensive stuff if this module has already been added
-    (( _addedmodules["$target"] > 0 )) && return
+    _add_modules=()
```
falconindy, line 384: Looks to be unused.

mkinitcpio
```diff
@@ -16,16 +16,16 @@ _d_hooks="$PWD/hooks:/usr/lib/initcpio/hooks:/lib/initcpio/hooks"
 _d_install="$PWD/install:/usr/lib/initcpio/install:/lib/initcpio/install"
 _d_flag_hooks=
 _d_flag_install=
-_d_firmware=({/usr,}/lib/firmware/updates {/usr,}/lib/firmware)
+_d_firmware=/usr/lib/firmware
```
falconindy, line 19: This is broken for any system that hasn't merged /lib and /usr/lib. The canonical location for firmware is /lib/firmware, not /usr/lib/firmware.

functions
falconindy, line 384: Should be declared in mkinitcpio with the rest of the globals, not here.

functions
```diff
@@ -336,58 +336,47 @@ add_checked_modules() {
...
 }
 
-add_module() {
-    # Add a kernel module to the initcpio image. Dependencies will be
-    # discovered and added.
-    #   $1: module name
-
-    local target= module= softdeps= deps= field= value= firmware=()
-    local ign_errors=0 found=0
+_add_module() {
```
falconindy, line 379: Seems like this should be plural since it potentially handles many arguments.

functions
```diff
@@ -397,24 +386,47 @@ add_module() {
...
+        return
     fi
 
-    # handle module quirks
-    case $target in
+    _add_module "${_add_modules[@]}"
+}
+
+# add_module adds a kernel modules $@ and all it's dependencies and firmware
+# to the image. ?-suffix ignores errors.
```
falconindy, line 438: add_module adds $@ as kernel modules with dependencies and firmware to the image. A suffix of '?' causes errors to be ignored.

functions
```diff
@@ -397,24 +386,47 @@ add_module() {
...
+    [[ $KERNELVERSION == none ]] && return 0
+
+    for target; do
+        target=${target%.ko*}
+        target=${target//-/_}
+
+        mod=${target%\?}
+
+        (( _addmodule[$mod] )) && continue
+        _addmodule[$mod]=1
```
falconindy, line 451: I think I understand why you have this, but it took me a solid 5-10 minutes to figure it out because the naming lead me to believe that it was a typo. Please drop it, or if you can show significant benetfit, add a comment explaining why we want the second layer of caching.

functions
```diff
@@ -428,7 +440,10 @@ add_module() {
         ext4)
             add_module "crypto-crc32c?"
```
falconindy, line 468: Why bother recursing? Just inline this as `_add_module+=(...)`. Don't forget to normalize `nls_iso8859-1`.

functions
```diff
@@ -390,6 +390,7 @@ _add_module() {
                 ;;
             filename)
                 quiet "filename: %s" "$value"
+                [[ "$value" == '(builtin)' ]] && continue
```
falconindy, line 393: Because this is a string madeup by kmod, I'm disinclined to depend on it. Can't we check that it looks like a path (similar to what I did)?

functions
```diff
@@ -336,58 +336,47 @@ add_checked_modules() {
...
             filename)
-                found=1
+                quiet "filename: %s" "$value"
                 module=${value##*/} module=${module%.ko*}
-                quiet "adding module: %s" "$module"
                 _modpaths["$value"]=1
                 _addedmodules["${module//-/_}"]=1
                 ;;
+            alias)
+                _addedmodules["${value//-/_}"]=3
```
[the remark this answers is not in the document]
tlahdekorpi: The only place where this is used outside of module resolution is parse_config() where it won't add builtins, and now aliases to MODULES. Currently this is not used anywhere to detect aliases so it could be changed to 2.

functions
tlahdekorpi: The amount of modules modinfo needs to resolve with this change drops in half and reduces the time spent in modinfo significantly.
```sh
# without
$ time ./mkinitcpio -c /dev/null -A block,filesystems,keyboard &> /dev/null

real	0m4.062s
user	0m3.873s
sys	0m1.566s

# with
$ time ./mkinitcpio -c /dev/null -A block,filesystems,keyboard &> /dev/null

real	0m2.423s
user	0m2.214s
sys	0m0.871s
```

functions
tlahdekorpi: Recursing here so that the above logic skips potentially duplicate modules.

functions
tlahdekorpi: Because (builtin) is a madeup string kmod sticks to builtin modules now, we can rely on builtin modules having it and skipping them. If kmod reports a non-path like string, we should not work around it.

falconindy: My point is that it might change in the future. Or, maybe kmod adds some other new madeup string to ascribe to some other kind of pseudo-module. In that case, we'd have to change our current exception or add a new one. Why bother trying to add things that don't look like paths?

tlahdekorpi: If kmod breaks this in the future, we can start checking for path-like strings. For now this is the only exception. If the filename does not look like a path outside of this, the bug is in the wrongly reported filename.

mkinitcpio
Maryse47: Does Arch Linux supports unmerged /lib?

grazzolini:
@tlahdekorpi I did some testing on this, and, although it seems to improve the time spent on module resolution, I found a bug that was hard to reproduce that would cause the module ordering to not be deterministic, causing the final image to not be reproducible. Can you rebase this and address the open threads so we can get things moving further on this one?

grazzolini:
I'm closing this PR. It has been a long time now and the PR was not rebased and also the original bug was fixed.

PR #23, opened 2020-05-20
Remove hexdump usage

E5ten:
lsinitcpio: Use read to get the first 6 bytes, convert them to hex, and
compare them to the hex versions of the magic bytes of each compression
format
Replace the kernel version offset determination using hexdump with dd to
skip bytes, and printf to convert the bytes to decimal

### DIFF
--- a/functions
+++ b/functions
@@ -144,7 +144,9 @@ kver() {
     # to the version string by reading 2 bytes out of image at at
     # address 0x20E. this leads us to a string of, at most, 128 bytes.
     # read the first word from this string as the kernel version.
-    local offset=$(hexdump -s 526 -n 2 -e '"%0d"' "$1")
+    local offset=$(dd if="$1" bs=1 count=2 skip=526 2>/dev/null)
+    printf -v offset '%02x' "'${offset:1:1}" "'${offset:0:1}"
+    printf -v offset '%d' "0x$offset"
     [[ $offset = +([0-9]) ]] || return 1
 
     read kver _ < \
--- a/lsinitcpio
+++ b/lsinitcpio
@@ -77,71 +77,51 @@ size_to_human() {
 detect_filetype() {
     local bytes
 
-    read -rd '' bytes < <(hexdump -n 6 -e '"%c"' "$1")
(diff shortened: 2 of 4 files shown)


falconindy:
Why do we care about removing hexdump?

E5ten:
In terms of the lsinitcpio change, hexdump is called 6 times with separate comparisons after each call, it seems a lot simpler to me to turn that into one read, a printf, and then one case for all the comparisons. For the other change (the same one in functions and completion files), dd is already used to find the kernel version using the given offset, so since it's possible to also obtain that offset using dd, which is a POSIX command, it makes sense to me to do that, instead of using 2 different tools for finding the version.

grazzolini:
@E5ten can you rebase this?

E5ten:
@grazzolini done

grazzolini:
@E5ten thanks. I'll do some testing on this. Not sure yet if it can make to the next release, but perhaps to the next one.

evelikov:
Some food for though - `detect_filetype` is effectively* only needed for the compression stats aka:
```
Compressed with: foo
Uncompressed size: 100 (0.007 ratio)
Estimated decompression time: 100s
```
Of which the estimate is off since it compares standalone decompressor + libarchive tar handling wrt the in-kernel code. One could remove it - or the whole section all together?

*Effectively - since in every other case, `bsdtar` can be used directly to decompress fully.

tlahdekorpi:
Like hexdump, dd also does not exit with a non-zero status code when you try to read past the selected offset. Using od to read and convert the offset to a number would be a better solution.

tlahdekorpi/mkinitcpio-ag@45285b4

nl6720 (changes requested):
This needs to be rebased after db25bf84f3d8740921bcd9035c9050b85b320a45.
And as suggested in https://github.com/archlinux/mkinitcpio/pull/23#issuecomment-777645680, changed to use `od`.

klausenbusk:
Moved to to: https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/merge_requests/23.

PR #24, opened 2020-05-20
install/strip: Replace usage of file with readelf

E5ten:


polyzen:
Please include the rationale in the commit body.

E5ten:
Done.

grazzolini:
@E5ten I don't see a rationale for this. We are swapping one dep for another here. What's the gain on doing this?

E5ten:
@grazzolini strip and readelf almost always come together, as part of binutils, or a similar package (e.g. elfutils, elftoolchain), so realistically, this is swapping 2 deps (file, and binutils for strip), for one dep (binutils for both strip and readelf).

grazzolini:
My point is, file is already part of base on Arch Linux. That's why it's used and not on the dep list of mkinitcpio.

E5ten:
Sure, but the purpose of this PR isn't to decrease the arch package's deps or something, it's just to decrease the amount of stuff mkinitcpio itself relies upon. This part of mkinitcpio already relies on some binutils-like package for the strip command, so imo it makes sense to use a command that also comes with any binutils-like package for the file type identification here, instead of a separate, unrelated command.

E5ten:
Now that binutils has been made an explicit dependency of mkinitcpio in https://github.com/archlinux/mkinitcpio/commit/b4fafb87ff0f9b25160d8954ef29bb39f958cb22, I feel that replacing `file` with a command that comes from binutils is even more justified than before.

grazzolini:
Can you update your PR? x-pie-executable was added, I don't know if it's covering that.

E5ten:
As far as I know, the correct way to handle PIE executables in regards to stripping is like shared libraries, which the PR still does. If I'm wrong about that, and they should be handled like non-PIE executables, I'll look further into updating to handle that.

E5ten:
https://gitlab.archlinux.org/pacman/pacman/-/commit/03272ad57142a1c7dacf1d9933d52650d3936225 the decision to use --strip-unneeded for PIE in makepkg was established here, for reference.

klausenbusk:
Moved to to: https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/merge_requests/24.

PR #25, opened 2020-05-20
do not uncompress and re-depmod kernel modules

tlahdekorpi:
Installing kernel modules as-is without uncompressing reduces time spent
generating the image significantly at the cost of a small size increase.

### DIFF
--- a/functions
+++ b/functions
@@ -769,8 +769,7 @@ try_enable_color() {
 }
 
 install_modules() {
-    local m moduledest=$BUILDROOT/lib/modules/$KERNELVERSION
-    local -a xz_comp gz_comp
+    local m
 
     [[ $KERNELVERSION == none ]] && return 0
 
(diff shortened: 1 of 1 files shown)


functions
```diff
@@ -779,36 +778,12 @@ install_modules() {
...
-
-    msg "Generating module dependencies"
-    install -m644 -t "$moduledest" "$_d_kmoduledir"/modules.builtin
-
-    # we install all modules into kernel/, making the .order file incorrect for
-    # the module tree. munge it, so that we have an accurate index. This avoids
-    # some rare and subtle issues with module loading choices when an alias
-    # resolves to multiple modules, only one of which can claim a device.
-    awk -F'/' '{ print "kernel/" $NF }' \
-        "$_d_kmoduledir"/modules.order >"$moduledest/modules.order"
```
falconindy, line 806: So we don't include the order file anymore? That's broken.

tlahdekorpi: modules.order is not used by modprobe

functions
```diff
@@ -769,8 +769,7 @@ try_enable_color() {
 }
 
 install_modules() {
-    local m moduledest=$BUILDROOT/lib/modules/$KERNELVERSION
-    local -a xz_comp gz_comp
+    local m
```
grazzolini, line 772: Can you make this variable more descriptive?

grazzolini:
@tlahdekorpi do you have some measurements of the improvements of image generation time for this? I confess I don't see the gain in doing this. we decompress the kernel modules for a reason.

grazzolini:
This had no movement in over a year. Closing.

PR #28, opened 2020-05-20
resolve symlinks when checking for deprecation

tlahdekorpi:
If a deprecation symlink points to a hook that collides with something
in PATH, it gets sourced instead. Use absolute paths when sourcing
install hooks.
```sh
$ readlink install/test
sleep
$ ./mkinitcpio -k none -c /dev/null -A test
==> Starting dry run: none
==> WARNING: Hook 'test' is deprecated. Replace it with 'sleep' in your config
functions: line 729: .: /usr/sbin/sleep: cannot execute binary file
==> ERROR: Failed to read sleep
==> Dry run complete, use -g IMAGE to generate a real image
```

### DIFF
--- a/functions
+++ b/functions
@@ -718,7 +718,7 @@ run_build_hook() {
     fi
 
     # check for deprecation
-    if resolved=$(readlink "$script") && [[ ${script##*/} != "${resolved##*/}" ]]; then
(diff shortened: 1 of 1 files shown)


grazzolini:
Merging #48 in. -e is stricter than -f.

PR #30, opened 2020-05-30
Add 'nvdimm' hook

orangecms:
This allows for booting from an ISO that was copied to a pmem device,
such as '/dev/pmem0'. See also https://github.com/u-root/webboot/.

Corresponding feature request: https://bugs.archlinux.org/task/66804

### DIFF
--- a/install/nvdimm
+++ b/install/nvdimm
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+build() {
+    add_checked_modules '/drivers/nvdimm/'
+}
+
+help() {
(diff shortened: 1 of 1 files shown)


nl6720:
Why not just add the modules to the `block` hook?

orangecms:
Either way works. It might be that these modules are not applicable for every user, so I gave them a separate hook. Could you file another proposal? Then upstream can decide which variant they prefer, if they accept it at all, that is. :)

grazzolini:
This is better implemented by #34

PR #35, opened 2020-07-26
mkinitcpio: Add support for the zstd compressor

luca020400:
This mainly for testing purposes until https://lkml.org/lkml/2020/7/23/1010 is merged in mainline.
Then this patch can  be safely applied

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -211,6 +211,9 @@ build_image() {
         lz4)
             COMPRESSION_OPTIONS+=('-l')
             ;;
+        zstd)
+            COMPRESSION_OPTIONS+=('-19')
+            ;;
     esac
 
     pushd "$BUILDROOT" >/dev/null
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -212,7 +212,7 @@ build_image() {
             COMPRESSION_OPTIONS+=('-l')
             ;;
         zstd)
-            COMPRESSION_OPTIONS+=('-19')
+            COMPRESSION_OPTIONS+=('-19' '-T0')
             ;;
     esac
 
--- a/mkinitcpio.conf
+++ b/mkinitcpio.conf
@@ -60,6 +60,7 @@ HOOKS=(base udev autodetect modconf block filesystems keyboard fsck)
 #COMPRESSION="xz"
 #COMPRESSION="lzop"
 #COMPRESSION="lz4"
+#COMPRESSION="zstd"
 
 # COMPRESSION_OPTIONS
 # Additional options for the compressor
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -15,6 +15,7 @@ optdepends=('xz: Use lzma or xz compression for the initramfs image'
             'bzip2: Use bzip2 compression for the initramfs image'
             'lzop: Use lzo compression for the initramfs image'
             'lz4: Use lz4 compression for the initramfs image'
+            'zstd: Use zstd compression for the initramfs image'
             'mkinitcpio-nfs-utils: Support for root filesystem on NFS')
 backup=(etc/mkinitcpio.conf)
 
--- a/lsinitcpio
+++ b/lsinitcpio
@@ -113,6 +113,10 @@ detect_filetype() {
             echo 'lz4 -l'
             return
             ;;
+        fd2fb528)
+            echo 'zstd'
+            return
+            ;;
     esac
 
     read -rd '' bytes < <(hexdump -n 3 -e '"%c"' "$1")
--- a/man/mkinitcpio.conf.5.txt
+++ b/man/mkinitcpio.conf.5.txt
@@ -54,8 +54,10 @@ Variables
 
 	Defines a program to filter the generated image through. As of linux 2.6.38,
 	the kernel understands the compression formats yielded by the *gzip*, *bzip2*,
-	*lz4*, *lzop*, *lzma*, and *xz* compressors. If unspecified, this setting
-	defaults to *gzip* compression. In order to create an uncompressed image, define
+	*lz4*, *lzop*, *lzma*, and *xz* compressors. As of linux 5.9 the kernel also
+	understands the compression formats yielded by the *zstd* compressor.
+	If unspecified, this setting defaults to *gzip* compression.
+	In order to create an uncompressed image, define
 	this variable as *cat*.
 +
 It's not hard to realize that a filter such as a *tac* or *rev* will cause


mkinitcpio
```diff
@@ -211,6 +211,9 @@ build_image() {
         lz4)
             COMPRESSION_OPTIONS+=('-l')
             ;;
+        zstd)
+            COMPRESSION_OPTIONS+=('-22' '--ultra')
```
luca020400, line 215: the options used are the same as found in the kernel proposal for the vmlinux image

luca020400:
For anyone interested testing
I backported the patches here: https://github.com/luca020400/linux
Apply this diff to the config https://gist.github.com/6631c82f4ed0c76267a1288376c34f71

luca020400:
@grazzolini Mind taking a look?

I uploaded the package to aur as mknitcpio-zstd ( along a supported kernel linux-llvm )

grazzolini:
Hi @luca020400. I plan on adding zstd support to mkinitcpio as soon as the kernel supports it. I also planned on dropping lz4 support, since it's the only compression algorithm that's not reproducible. If the Arch kernel got patched to support zstd, I might consider patching the mkinitcpio package as well, but that's not something I'm considering pushing to the kernel maintainers currently.

luca020400:
Feel free to use these commits, they've been tested.
zstd support might come in 5.9, so I guess we can wait.
As seen here https://lkml.org/lkml/2020/7/24/677 bzip2 and lzma are likely going to be dropped from kernel, so we might want to remove them as well.
Didn't know lz4 isn't reproducible, but honestly lz4 is a good algorithm extensively used thanks to the quick decompression, I've been using it before switching to zstd.
I think as soon as zstd lands lz4 removal is doable, and actually switching the default

grazzolini:
I was told that at least one of the officially supported kernels for Arch, linux-zen, has zstd support, due to the fact that the zen patchset adds zstd support. I'll give your patches some testing and try to merge them and make a mkinitcpio release before 5.9, that is, considering 5.9 really adds zstd support.

luca020400:
I hope it does, but all I can do is adding by "Tested-by" sign-off on the changes. I'd really want to see zstd being in.
Didn't know linux-zen support zstd compression, then I think you can safely add this feature even now.

falconindy:
> then I think you can safely add this feature even now.
We have no checks in mkinitcpio for examining compression support. If a user on a non-zen kernel tries to produce a zstd initramfs, they're going to have a bad time.

I *really* don't see any rush for getting this into mkinitcpio mainline. Nothing about the current codebase prevents you (an interested early adopter) from using zstd on your own kernel.

luca020400:
@falconindy agree, then we shall wait for mainline adoption of zstd

grazzolini:
Well, it *can* land on mkinitcpio before that. I'm particular ok with adding zstd support but not adding it as default on mkinitcpio.conf. Perhaps not even list it there for now. At least one official kernel (-zen) supports it already. When zstd lands on mainline, we can add zstd to the default .conf file. And, after a lot of testing (and giving people time and warning), perhaps even change it to default.

luca020400:
@grazzolini feel free to amend my commit to remove it from the default .conf if you want to do that.
I'd be happy :)

eworm-de:
You should update the info in man page... Support for zstd was not added in linux 2.6.38.

mkinitcpio
grazzolini: You have a link for this?

luca020400: https://lkml.org/lkml/2020/7/23/1013
But I've actually read the documentation and -22 shouldn't be used, instead for initramfs -19 should do the job

luca020400:
Incorporated changes made by @eworm-de for man page
Also fixed compression options to use -19

mkinitcpio.conf
```diff
@@ -60,6 +60,7 @@ HOOKS=(base udev autodetect modconf block filesystems keyboard fsck)
 #COMPRESSION="xz"
 #COMPRESSION="lzop"
 #COMPRESSION="lz4"
+#COMPRESSION="zstd"
```
grazzolini, line 63: I'm still wary of adding this to the .conf file, even commented, because the users might get the impression that we support zstd, which we don't yet. I was thinking about reaching to the Arch kernel maintainers, @heftig, @anthraxx, and others, to see if we could have the zstd patch merged across the board, so this wouldn't be an issue.

eworm-de:
Linux 5.8 hast just been released, linux 5.10 will be in about four month. That is when linux 5.9 moves to linux-lts package. So that is latest date to release full support.

mkinitcpio
```diff
@@ -211,6 +211,9 @@ build_image() {
         lz4)
             COMPRESSION_OPTIONS+=('-l')
             ;;
+        zstd)
+            COMPRESSION_OPTIONS+=('-19')
```
codicodi, line 215: How about  `-T0` (multithreaded compression)?

luca020400: None of the tools used in mkinitcpio are threaded, so probably not

grazzolini: If -T0 is faster (should be) and still makes reproducible archives, that's great. On the other hand, I'd say we should use reasonable conservative defaults (matching upstream zstd defaults), and let users tweak further, if they want.

codicodi: It should be reproducible [1]. `-T0` is also used in devtools [2], but admittedly, unlike some packages, initcpios aren't usually large enough for parallel compression to make a big difference.

[1] https://github.com/facebook/zstd/issues/2238
[2] https://github.com/archlinux/devtools/blob/20200407/makepkg-x86_64.conf#L135

Maryse47: > I'd say we should use reasonable conservative defaults (matching upstream zstd defaults)

I think `-19` isn't upstream default already otherwise you wouldn't need to pass it explicitly.

klausenbusk: Debian uses `-19 -T0` for what it's worth: https://salsa.debian.org/kernel-team/initramfs-tools/-/blob/cbbbb1b73e72124ba517325c51e0ea4bd7139104/mkinitramfs#L187

luca020400: They also use pigz for gzip is available, we don't.
Maybe we can just do another PR to switch to multi-threaded tools if possibile?

grazzolini: I'm fine with using -T0 right now, as long as it's reproducible (it seems to be). Also, what I meant by -19 upstream is the kernel recommendation to use -19, not zstd's. Anyway, @luca020400, if you want to update this branch to add -T0, I'm fine with it.

luca020400: Done

heftig: Will it even matter in this case? What's the block size used for -19 compared to the size of the typical initramfs?

codicodi: I think you mean window size? It's 21M. Fallback initramfs can easily be 100+. 
On my machine `-T0` speeds up `mkinitcpio -P` by 35%.

grazzolini: The regular initramfs is around that window, but I'd say it still has benefits, even if you, like me, aren't using the fallback initramfs (I'm thinking about disabling it's generation *by default*)

luca020400:
Seems like @heftig defaulted 5.9 to ZSTD ( which also brings in the CONFIG_RD_ZSTD needed for ramdisk )
Zen already backported them, I guess we have to wait for LTS now

grazzolini:
I won't wait until LTS gets support for this. Now that 5.9 has zstd support, I'll merge this PR and make a new mkinitcpio release (after a ton of testing). If someone tries to use zstd with LTS, that's their fault, because the documentation will state accurately which kernel got zstd support. Nor mkinitcpio will switch to zstd by default, in the foreseeable future.

luca020400:
Good for me then. Ship it

eworm-de:
Greg fooled us... Linux 5.9 is not going to be LTS - but 5.10 will. So we will have to wait for two more month for zstd compressed initcpio with linux-lts.

grazzolini:
Well, I've done a few tests already with 5.9, it works fine. I'm cleaning PR's this week and making a release. But anyone willing to use zstd can do so already, only lsinitcpio won't (properly) work on the images.

grazzolini (approved):

PR #37, opened 2020-08-14
support zstd compression

eworm-de:
Linux 5.9 supports zstd compressed initramfs, so add support here.


eworm-de:
Well, #35 is pretty incomplete... In fact it just sets compression options - which my commit does not - but lacks everything else.
Do we want to set a compression level of is the user supposed to choose one?

grazzolini:
I'm going to probably merge and test both your PR and #35. Also, see my latest comment there. Anyway, -zen already supports zstd too, so this could go in already. But, not as default, for now.

eworm-de:
Closing as #35 has merged all my changes.
Thanks!

PR #38, opened 2020-10-20
Fix setting uid=0 gid=0 in archive

arkdchst:
Setting uid and gid in bsdtar (step 2) gave no effect, so it's moved to step 1. Now all uids and gids are setting to zero. So created archive is user-independent.
Step 2 just adds files from archive from stdin, but doesn't set any metadata to them.
Before that if you ran mkinitcpio as non-root user, uid and gid of each file from archive were set to your non-zero uid and non-zero gid.

**This bug caused a following problem:**
If you create initramfs with non-root user and try to boot with it, you will get
```
mount: /proc: must be superuser to use mount.
mount: /sys: must be superuser to use mount.
mount: /dev: must be superuser to use mount.
mount: /run: must be superuser to use mount.
```
because all files in your initrd environment have non-zero uid and gid.

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -221,8 +221,8 @@ build_image() {
     # If this pipeline changes, |pipeprogs| below needs to be updated as well.
     find . -mindepth 1 -printf '%P\0' |
             sort -z |
-            LANG=C bsdtar --null -cnf - -T - |
-            LANG=C bsdtar --uid 0 --gid 0 --null -cf - --format=newc @- |
+            LANG=C bsdtar --uid 0 --gid 0 --null -cnf - -T - |
+            LANG=C bsdtar --null -cf - --format=newc @- |
             $compress "${COMPRESSION_OPTIONS[@]}" > "$out"
 
     pipestatus=("${PIPESTATUS[@]}")


Foxboron:
Does this affect the reproducability of the images?

arkdchst:
This fully saves current level of reproducibility, but also adds user-independence.

arkdchst:
I also opened bug report here: https://bugs.archlinux.org/task/68349

Foxboron:
Which images are from which versions. It's a bit unclear from your comment. Is `4.img` from the release or after your commit?

arkdchst:
@Foxboron 
`4.img` is from the release.
Ok, I think it would be better:
```bash
git checkout 824f35b #my commit
./mkinitcpio -g ./my.img
sudo ./mkinitcpio -g ./my-sudo.img

git checkout 824f35b~ #previous commit
./mkinitcpio -g ./prev.img
sudo ./mkinitcpio -g ./prev-sudo.img

git checkout v27 #release
./mkinitcpio -g ./release.img
sudo ./mkinitcpio -g ./release-sudo.img

sha1sum *img
```
After that I receive this:
```
e0d90c86185e06925cb981d1d0e5d028d6688bdd  my-sudo.img
e0d90c86185e06925cb981d1d0e5d028d6688bdd  my.img
e0d90c86185e06925cb981d1d0e5d028d6688bdd  prev-sudo.img
ce696ad44e605997c43fade78490a771585023a3  prev.img
e0d90c86185e06925cb981d1d0e5d028d6688bdd  release-sudo.img
ce696ad44e605997c43fade78490a771585023a3  release.img
```

Foxboron:
Ah yes, that is much better! Thanks :) But that seems like my concerns has been addressed. So LGTM.

arkdchst:
@Foxboron,
Great! Will it be merged?

grazzolini:
@arkdchst I'll have to confirm this doesn't break reproducibility, because we specifically changed the order to make it reproducible.

grazzolini:
@arkdchst This does seem to break reproducibility on initial testing I've done. Can you re-work this in a way that doesn't?

arkdchst:
Can you give me please environment in which reproducibility breaks or tests you have done?

bartsch:
Please, merge the commit. Just hit this bug too.

tcodes0:
This needs to be merged, everything in the wiki suggests running mkinitcpio as a regular user

PR #39, opened 2020-10-30
Allow user to override default compression options

Chocobo1:
Background info: I was using lz4 for its speed and switched to zstd after the recent release. However `zstd -19` is very slow on my machine and I wish to override the compression level which I found is impossible and hence this patch.

Another related question, why only zstd utilize a [compression level](https://github.com/archlinux/mkinitcpio/blob/master/mkinitcpio#L215) that is not the program default (`-3`)? All others (lz4, gzip) use the default compression level from the respective programs.
EDIT: Answering the question myself, see: https://github.com/archlinux/mkinitcpio/pull/35#discussion_r460510753

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -206,13 +206,13 @@ build_image() {
             msg "Creating %s-compressed initcpio image: %s" "$compress" "$out"
             ;;&
         xz)
-            COMPRESSION_OPTIONS+=('--check=crc32')
+            COMPRESSION_OPTIONS=('--check=crc32' "${COMPRESSION_OPTIONS[@]}")
             ;;
         lz4)
-            COMPRESSION_OPTIONS+=('-l')
+            COMPRESSION_OPTIONS=('-l' "${COMPRESSION_OPTIONS[@]}")
             ;;
         zstd)
-            COMPRESSION_OPTIONS+=('-19' '-T0')
+            COMPRESSION_OPTIONS=('-19' '-T0' "${COMPRESSION_OPTIONS[@]}")
             ;;
     esac
 


pfactum:
https://bugs.archlinux.org/task/68470

Chocobo1:
Hmm, (I could be wrong) maybe the compression levels are not meant to be override freely...

pfactum:
> Hmm, (I could be wrong) maybe the compression levels are not meant to be override freely...

It should be allowed. Not all the hardware is happy dealing with `-19` compression level by default. Even my Zen 2 struggles with it, default gzip is way faster. So IMO zstd should default to `-10` or even `-3`, but there should be a possibility to override it anyway.

Maryse47:
> Another related question, why only zstd utilize a compression level that is not the program default (-3)? All others (lz4, gzip) use the default compression level from the respective programs.
> EDIT: Answering the question myself, see: #35 (comment)

I think that answer is incomplete. Linked kernel doc only says that `-22` shouldn't be used and it doesn't says that `-19` should. It just provides it as example. It also provides `--lzma2=dict=1MiB` as example for xz but mkinitcpio doesn't care then why suddenly `-19` is treated as mandatory especially if it's proven to suck even on beefy hardware and it's impossible to change it for users.

Foxboron:
I think adding `ZSTD_COMPRESSION_OPTIONS` and the likes would be a better approach. Then you include recommended, or have default, options without the fuzz of sharing one variable.

@grazzolini thoughts?

grazzolini:
Whatever options are appended to zstd seems to take precedence. I'm not sure if it's possible to disable -T0 with a subsequent flag. So, basically, the options coming from mkinitcpio.conf should be appended to defaults. This patch seems to be doing that, but I'm not sure that's the case for all compression tools (some might ignore further flags and consider only the first).

Chocobo1:
>So, basically, the options coming from mkinitcpio.conf should be appended to defaults. This patch seems to be doing that, but I'm not sure that's the case for all compression tools (some might ignore further flags and consider only the first).

I tested both xz and lz4 and latter options are taking precedence as expected.

>I'm not sure if it's possible to disable -T0 with a subsequent flag.

Might be possible by appending a `-T1` option, I didn't test that though.

rpthms:
For now, I'm just setting COMPRESSION to "zstdmt" instead of "zstd" to avoid the -19. According to the man page, zstdmt is equivalent to zstd -T0.

grazzolini:
I plan on making a new mkinitcpio release this year still, but let's see how things go.

PR #43, opened 2021-01-08
Add zstd module decompression

sirlucjan:
kmod 28 supports modules compressed in zstd format so let's add the same functionality to mkinitcpio.

Signed-off-by: Piotr Gorski <lucjan.lucjanov@gmail.com>

### DIFF
--- a/functions
+++ b/functions
@@ -803,7 +803,7 @@ install_modules() {
     done
     (( ${#xz_comp[*]} )) && xz -d "${xz_comp[@]}"
     (( ${#gz_comp[*]} )) && gzip -d "${gz_comp[@]}"
-    (( ${#zst_comp[*]} )) && zstd -d -q "${zst_comp[@]}"
+    (( ${#zst_comp[*]} )) && zstd -d --rm -q "${zst_comp[@]}"
 
     msg "Generating module dependencies"
     install -m644 -t "$moduledest" "$_d_kmoduledir"/modules.builtin
--- a/functions
+++ b/functions
@@ -776,7 +776,7 @@ try_enable_color() {
 
 install_modules() {
     local m moduledest=$BUILDROOT/lib/modules/$KERNELVERSION
-    local -a xz_comp gz_comp
+    local -a xz_comp gz_comp zst_comp
 
     [[ $KERNELVERSION == none ]] && return 0
 
@@ -796,10 +796,14 @@ install_modules() {
             *.gz)
                 gz_comp+=("$moduledest/kernel/${m##*/}")
                 ;;
+            *.zst)
+                zst_comp+=("$moduledest/kernel/${m##*/}")
+                ;;
         esac
     done
     (( ${#xz_comp[*]} )) && xz -d "${xz_comp[@]}"
     (( ${#gz_comp[*]} )) && gzip -d "${gz_comp[@]}"
+    (( ${#zst_comp[*]} )) && zstd -d "${zst_comp[@]}"
 
     msg "Generating module dependencies"
     install -m644 -t "$moduledest" "$_d_kmoduledir"/modules.builtin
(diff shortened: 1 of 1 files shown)


nycko123:
after several weeks, works with no failures - locally compiled mkinitcpio-zstd

pacman -Qs "kmod|mkinitcpio"
> local/kmod 28-1
local/mkinitcpio-busybox 1.32.1-1
local/mkinitcpio-zstd 29-6

/etc/mkinitcpio.conf
>MODULES=(i915)
BINARIES=()
FILES=(/etc/vconsole.conf)
HOOKS=(base udev autodetect keyboard modconf block encrypt filesystems fsck)
COMPRESSION="zstd"
COMPRESSION_OPTIONS=('-10' '-T8' '-c' '-z' '-q')

mkinitcpio -p customkernel
[log file](https://paste.rs/6Zp)

lingruby-pl:
```
pacman -Qs "kmod|mkinitcpio"


local/kmod-zstd 27-4
    Linux kernel module management tools and library
local/kmod-zstd-debug 27-4
    Detached debugging symbols for kmod-zstd
local/mkinitcpio-zstd 29-6
    Modular initramfs image creation utility
```
/etc/mkinitcpio.conf

```
MODULES=(ext4)
HOOKS=(base)
COMPRESSION="zstd"
COMPRESSION_OPTIONS=('-10' '-T8' '-c' '-z' '-q')
```

works correctly

functions
```diff
@@ -796,10 +796,14 @@ install_modules() {
...
                 gz_comp+=("$moduledest/kernel/${m##*/}")
                 ;;
+            *.zst)
+                zst_comp+=("$moduledest/kernel/${m##*/}")
+                ;;
         esac
     done
     (( ${#xz_comp[*]} )) && xz -d "${xz_comp[@]}"
     (( ${#gz_comp[*]} )) && gzip -d "${gz_comp[@]}"
+    (( ${#zst_comp[*]} )) && zstd -d "${zst_comp[@]}"
```
pfactum, line 806: I believe `zstd` should be silenced here with the `-q` option in order to avoid garbage during initramfs regeneration both in console and pacman log.

sirlucjan:
Thanks for the suggestion, done.

functions
sirlucjan: You're absolutely right, I overlooked that.

zhmars:
Correct me if I'm wrong. I believe the `--rm` option is also needed here to remove  source module files, like `xz/gzip` does.

sirlucjan:
I'm pretty sure it's not necessary in this case. It's totally necessary when compressing files but in this case I think it happens automatically. 

@pfactum What do you think about this? 

Adding this will not be the slightest problem.

zhmars:
>  -k, --keep: keep source file(s) after successful compression or decompression. This is the default behavior.

It seems the source files are preserved by default during initramfs regeneration. 
With the `--rm` option, we can get a smaller initramfs file. (for me, 11M -> 8.5M)

sirlucjan:
Done.

sirlucjan:
@zhmars Done.

pfactum:
@sirlucjan agree with this, it seems that zstd has got inverted defaults unlike gzip and xz.
@zhmars thanks for noticing this.

pfactum:
@grazzolini please note we are pushing related work to the upstream kernel (https://github.com/hnaz/linux-mm/commit/bb63820634e41278399f3d65ca0f8306eca67351) so that this pull request is expected to be actually usable without kernel patching soon.

sirlucjan:
@pfactum @grazzolini 

The patch has been merged into linux-next so perhaps this will appear in 5.13

https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/commit/?id=7b35c1a01c2e2f04638312c4051cf30003dd406d

grazzolini:
I have just merged this. However, this doesn't mean Arch kernels will start compressing modules with zstd as soon as 5.13 lands.

PR #44, opened 2021-01-08
mount handler fallback shell

paledega:
If mount handler crash, we just see kernel panic kill init message. 

I added fallback shell. now we can see error message.

### DIFF
--- a/init
+++ b/init
@@ -41,7 +41,7 @@ unset rootdev
 fsck_root
 
 # Mount root at /new_root
-"$mount_handler" /new_root
+"$mount_handler" /new_root || launch_interactive_shell --exec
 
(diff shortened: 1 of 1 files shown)


paledega:
image

our hook:
https://gitlab.com/tearch-linux/tearch-live-hooks/-/blob/master/hook_live_hook

grazzolini:
The thing is, on mkinitcpio we have only the default_mount_handler. And that one already spawns a shell in case root was not able to be mounted. I suggest you mimic the default_mount_handler behavior withing your own hook and spawn the interactive shell from there. That way you have better control.

Foxboron:
I don't think this belongs in mkinitcpio really and should maybe be a seperate hook.

PR #45, opened 2021-01-16
mkinitcpio: Support kernel version detection on other archtectures

adam900710:
Current mkinitpcio only support kernel version detection on x86/x86_64.

This makes mkinitcpio on Archlinuxarm/ManajaroARM harder to generate their
initramfs, as they have to specify the kernel version, other than let
mkinitcpio to detect it automatically.

This limitation is especially obvious when kernel bisecting is involved,
as the kernel version can change between each bisect, and re-generating
initramfs with inconsistent kernel version is a blockage.

This patch adds a fallback method to detect kernel version for
uncompressed kernel image.
The fallback method will use "strings" command to filter out the
"Kernel version " string and use the next string as kernel version.

This works on arm64 and x86_64 uncompressed kernel pretty well, but it's
still slower since we need to read the full kernel image.

Thus the existing x86_64 magic offset way is still kept as is, and the
new generic method is only a fallback for arch without a magic offset.

Signed-off-by: Qu Wenruo <wqu@suse.com>

### DIFF
--- a/functions
+++ b/functions
@@ -139,6 +139,21 @@ kver() {
     # resembling dotted decimal notation. remember that there's no
     # requirement for CONFIG_LOCALVERSION to be set.
     local kver re='^[[:digit:]]+(\.[[:digit:]]+)+'
+    local arch=$(uname -m)
+
+    if [[ $arch != @(i?86|x86_64) ]]; then
+        # Use the generic way by searching the plaintext string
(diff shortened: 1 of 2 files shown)


adam900710:
Any feedback?

I really hope we can get this feature merged so that all arch based ARM distro could use much simplified preset.

grazzolini:
implemented by #32

PR #47, opened 2021-01-30
Change explicit default for compression level of zstd

cherti:
This PR removes the explicit default of -15 for zstd compression.

The reason for this proposal is, as can be seen in the plot below, that -15 is still paying significant CPU cost for barely any gain. By removing -15, zstd would, if the user didn't specify anything else, resort to level 3, which saves significant compression times while raising the filesize only by 2.5% of the original size.

Furthermore, this would make the behavior of mkinitcpio less surprising in this regard, as all other compressors are used with their default levels, only zstd has an explicit level set, potentially leading to user confusion, why zstd, known among other things for its compression speed, is slower than other compressors like lz4.

Given the data shown below and the principle of least surprise, and given that the default of zstd already performs equal in size to lz4 but noticably better in compression ratio, I think that `-15` should be lowered for the default choice and not explicitly setting a compression level (and hence using `zstd -3`) is a sane default here.

compress-stats-mkinitcpio

Notation: `compression-algorithm.level`

Methodology: The displayed values are median values of multiple runs, to reduce statistical flaking, the original sample was cached from disk to reduce the impact of disk-delay, and the system was run with only the Terminal open holding the benchmark process, to reduce noise from other applications. Standard background processes were not stopped, interference of those is considered to be dealt with by the multiple runs and median-taking afterwards.

Decompression remark: For decompression the level doesn't matter, `zstd` is so fast that I can't even say that the decompression times I get are decompression times and not dominated by things that are not zstd actually doing work (for all levels we are solidly below 0.1s for decompression time, and the decompression time is barely even systematically dependent on the compression level, so when thinking about the desired level, decompression time doesn't seem to bring anything to the table it terms of reasons for decisions).

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -212,7 +212,7 @@ build_image() {
             COMPRESSION_OPTIONS=('-l' "${COMPRESSION_OPTIONS[@]}")
             ;;
         zstd)
-            COMPRESSION_OPTIONS=('-15' '-T0' "${COMPRESSION_OPTIONS[@]}")
+            COMPRESSION_OPTIONS=('-T0' "${COMPRESSION_OPTIONS[@]}")
             ;;
     esac
 


grazzolini:
Well, I did my comparison with gzip (which is the current default), but I got similar times as you did. Using -15 takes roughly the same time as gzip currently does, but it yields a smaller image. Using -3, will yield a bigger image, as you also demonstrated. I plan to change to zstd by default, but I don't want to break people's boot because they *might* not have the space on their boot partition for the new default zstd generated images. Hence why I did *a lot* of experimentation (albeit, I didn't create a nice graphic like you did =D), and using -15 *for now* will allow us to switch to zstd by default, while making mkinitcpio to take roughly the same time gzip currently does, and produce smaller images.

cherti:
ah, I see, that does make sense regarding the reasoning for `-15` indeed, and I am very much liking the "for now".^^

Although, I want to remark:

> using -15 for now will allow us to switch to zstd by default, while making mkinitcpio to take roughly the same time gzip currently does, and produce smaller images.

Doesn't this lead to the problem that if zstd produces smaller images in the same time, this will mean that by your reasoning you cannot ever increase the default level without risking this very same breakage in the future?
I think by aiming at same file size, this problem could be resolved, hence I'm still a little wary about the `-15` here.

The primary reason for this is, while I do not doubt in the slightest that you have made proper measurements, I cannot really substantiate that with my numbers here:

with_gzip

This looks pretty much the same for a fallback-image of mkinitcpio, I of course have that same plot as well, if you are interested.

So if you intend to migrate to default-levels for zstd in the future, wouldn't it be less risky to do so right now, producing images of roughly the same size, but faster? (Which in turn would make the new mkinitcpio also much more noticable to the user by improved speeds.^^)

grazzolini:
Ideally we should get smaller images, faster. But, for example, we are already adding -T0 to zstd, so, we're not even using defaults. I did my measurements on 2 machines, but, I didn't use any script like you're using. Also, I restricted myself to regular images, not the fallback ones (which I plan to deprecate in the future). gzip's default is -6. Don't get me wrong, one of the reasons for adding zstd as default is its speed, but, on space constrained environments like boot, a few MB makes all the difference. So, I think that, changing to zstd by default and *at the same time* increasing the image size is not a good idea. If you want to use your script to try to find a value better than -15, that yields images *at least* same size as gzip does, I'm all ears.

cherti:
> If you want to use your script to try to find a value better than -15, that yields images at least same size as gzip does

So maybe we have a misunderstanding here, but at least from the numbers I have measured, `zstd -3` compresses my mkinitcpio image down to 34% of its original size, while `gzip -6` ends up compressing down to 36%, so `zstd` in its default configuration already yields smaller images than `gzip` does (in fact, `zstd` with it's default of `-3` yields smaller sizes `gzip` is even capable of producing, as far as my samples go).

So `zstd -3` does already yield a smaller image than gzip? If that was not what you were after, can you rephrase what you mean? :)

> You might also want to share it, so I can add it to the repo as a compression speed/size test.

That sounds like an interesting idea, I'll have to polish it a little more before, though, so that it works more autonomously.

grazzolini:
> 
> So maybe we have a misunderstanding here, but at least from the numbers I have measured, `zstd -3` compresses* my mkinitcpio image (from uncompressed) down to 34% of its original size, while `gzip -6` ends up compressing down to 36%, so `zstd` in its default configuration already yields smaller images than `gzip` does (in fact, `zstd` with it's default of `-3` yields smaller sizes `gzip` is even capable of producing, as far as my samples go).
> 
> So `zstd -3` does already yield a smaller image than `gzip`, meaning it would be an unconditional improvement. If I'm misunderstanding your point here, can you rephrase what you mean? :)
> 

Looking at both zstd code and gzip code, it seems you're right. Can you rebase this and drop the explicit -3?

> 
> That sounds like an interesting idea! I'll have to polish it a little more before, though, so that it works more autonomously.
> 

Sure, if you want to.

> *) when I say "compressed down", I mean I took the uncompressed image and threw it against the different compressors right on the command line, not through `mkinitcpio.conf`, but this shouldn't make a difference, as far as I can tell.

It shouldn't have any difference.

grazzolini (approved):

PR #48, opened 2021-02-08
functions: canonicalize symbolic link path

eworm-de:
Deprectating a hook 'old-hook' and replacing it with new hook 'new-hook'
required the symlink to point to an absolute target:

/usr/lib/initcpio/install/old-hook -> /usr/lib/initcpio/install/new-hook

Let's canonicalize the path read by `readlink`, this allows to link
relative path in symbolic link:

/usr/lib/initcpio/install/old-hook -> new-hook

### DIFF
--- a/functions
+++ b/functions
@@ -724,7 +724,7 @@ run_build_hook() {
     fi
 
     # check for deprecation
-    if resolved=$(readlink "$script") && [[ ${script##*/} != "${resolved##*/}" ]]; then
+    if resolved=$(readlink -e "$script") && [[ ${script##*/} != "${resolved##*/}" ]]; then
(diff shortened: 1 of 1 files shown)


eworm-de:
Had to work around this issue here: https://github.com/archlinux/svntogit-packages/commit/6e11efabb3ceab9e49cb1853a04429d6ca71a588

eworm-de:
On the other side... Why do we redefine `script`? Any reason not to read from symlink after printing the warning?

tlahdekorpi:
Duplicate of #28

As for the why, because add_runscript uses BASH_SOURCE to determine what to add. Not redefining would cause it to add the wrong runscript.

grazzolini (approved):

PR #49, opened 2021-02-09
functions: read from symlink, do not update script path

eworm-de:
Deprectating a hook 'old-hook' and replacing it with new hook 'new-hook'
required the symlink to point to an absolute target:

/usr/lib/initcpio/install/old-hook -> /usr/lib/initcpio/install/new-hook

Let's just print the warning, then read from symlink. This way we can not
break the path and a relative path works as well.

/usr/lib/initcpio/install/old-hook -> new-hook

### DIFF
--- a/functions
+++ b/functions
@@ -727,7 +727,6 @@ run_build_hook() {
     if resolved=$(readlink "$script") && [[ ${script##*/} != "${resolved##*/}" ]]; then
         warning "Hook '%s' is deprecated. Replace it with '%s' in your config" \
             "${script##*/}" "${resolved##*/}"
-        script=$resolved
     fi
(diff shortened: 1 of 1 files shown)


eworm-de:
This is an alternative to #48 ...

There's a little difference in output:
```
==> WARNING: Hook 'old-hook' is deprecated. Replace it with 'new-hook' in your config
  -> Running build hook: [old-hook]
```
vs.
```
==> WARNING: Hook 'old-hook' is deprecated. Replace it with 'new-hook' in your config
  -> Running build hook: [new-hook]
```

grazzolini:
Merged #48 in.

PR #53, opened 2021-04-05
[mkinitcpio] Create UEFI executables

Foxboron:
Missing stuff for this pull-request.

- [ ] Do we want a new section in the manpages?
- [ ] Is the flags okay enough?
- [x] Testing

---

Implement UEFI executable generation in mkinitcpio by utilizing UEFI
stubs provided by systemd/gummiboot.

This allows us to create a unified boot image we can boot from UEFI
with. These are practical for secure boot as we can sign initramfs,
kernel cmdline and the kernel all at once. By utilizing the
BOOT_LOADER_SPECIFICATION we can also drop new images into the correct
patch and have systemd-boot/gummiboot pick up the images.

The code does several things and does a fair amount of guessing to
figure out all the inputs needed.

We use `/etc/kernel/cmdline` to localize the kernel cmdline options we
want for the image. This is inherited from the `kernel-install` hook
system which might double as some form of standard.

We also do a dance to get the correct kernel image. We do a lookup into
/lib/modules and /boot for both versioned and unversioned kernels
(mainly Arch Linux).

There is an attempt to support both 32bit and 64bit lookup paths for the
stub images, but only 64bit is tested.

Gummiboot is also not tested.

https://www.freedesktop.org/software/systemd/man/kernel-install.html

https://systemd.io/BOOT_LOADER_SPECIFICATION/#type-2-efi-unified-kernel-images

### DIFF
--- a/man/mkinitcpio.8.txt
+++ b/man/mkinitcpio.8.txt
@@ -35,7 +35,7 @@ Options
 	or '/proc/cmdline'. Only used along with '--uefi'.
 
 *-c, \--config* 'config'::
-	Use 'config' file to generate the ramdisk. Default: '/etc/mkinitcpio.conf'
+	Use 'config' file to generate the ramdisk. Default: '/etc/mkinitcpio.conf'.
 
 *-d, \--generatedir* 'directory'::
 	Set 'directory' as the location where the initramfs is built. This might be
@@ -109,7 +109,7 @@ Options
 *-r, \--moduleroot* 'root'::
 	Specifies the root directory to find modules in, defaulting to '/'.
 
-*-R, \--osrelease* 'filename';;
+*-R, \--osrelease* 'filename'::
 	Include a os-release file for the UEFI executable. Only used if '--uefi' is
 	specified. Default: '/etc/os-release' or '/usr/lib/os-release'.
 
@@ -382,6 +382,15 @@ Examples
 *mkinitcpio -g /boot/initramfs-linux.img -k /boot/vmlinuz-linux*::
 	Create an initial ramdisk for the kernel at /boot/vmlinuz-linux. The
 	resulting image will be written to /boot/initramfs-linux.img.
+
+*mkinitcpio -g /boot/initramfs-linux.img -k /boot/vmlinuz-linux --uefi /efi/EFI/Linux/systemd-linux.efi*::
+	Create an initial ramdisk for the kernel along with a UEFI executable.
+	The resuling executable will be written to /efi/EFI/Linux/systemd-linux.efi.
+
+*mkinitcpio -g /boot/initramfs-linux.img -k /boot/vmlinuz-linux -U /efi/EFI/Linux/systemd-linux.efi -M /boot/intel-ucode.img -l /usr/share/systemd/bootctl/splash-arch.bmp*::
+	Create an initial ramdisk for the kernel and an UEFI executable. This
+	also includes the Intel CPU microcode and a splash image which will be
+	used during boot.
 
 See also
 --------
--- a/man/mkinitcpio.8.txt
+++ b/man/mkinitcpio.8.txt
@@ -29,11 +29,6 @@ Options
 	after all other hooks from the config file. Multiple hooks should be
 	comma-separated. This option can be specified multiple times.
 
-*-C, \--cmdline* 'config'::
-	Use kernel cmdline with UEFI executable. If none is specified it will
-	try find one of the files '/etc/kernel/cmdline', '/usr/share/kernel/cmdline'
-	or '/proc/cmdline'. Only used along with '--uefi'.
-
 *-c, \--config* 'config'::
 	Use 'config' file to generate the ramdisk. Default: '/etc/mkinitcpio.conf'.
 
@@ -64,28 +59,10 @@ Options
 *-L, \--listhooks*::
 	List all available hooks.
 
-*-l, \--splash* 'filename'::
-	UEFI executables can show a bitmap file on boot.
-
-*-I, \--uefistub* 'filename'::
-	UEFI stub image used for UEFI executable generation. Only used if '--uefi'
-	is specified. Default: Attempts to look for a systemd-boot or gummiboot
-	stub loader.
-
-*-i, \--kernelimage* 'filename'::
-	Include a kernel image for the UEFI executable. This is only used if
-	'--uefi' is specified.  Default: one of
-	'/lib/modules/$KERNELVERSION/vmlinuz', '/boot/vmlinuz-$KERNELVERSION', or
-	'/boot/vmlinuz-linux'.
-
 *-M, \--automods*::
 	Display modules found via autodetection. mkinitcpio will automatically try to
 	determine which kernel modules are needed to start your computer. This option
 	lists which modules were detected.
-
-*-m, \--microcode* 'filename'::
-	Include microcode into the UEFI executable. This option is only used if
-	'--uefi' is specified. Default: no.
 
 *-n, \--nocolor*::
 	Disable color output.
@@ -96,6 +73,7 @@ Options
 	booting. This combines the initramfs, the kernel, any specified microcode
 	and the kernel cmdline into one executable. This is useful for boot chain
 	integrity where the file is signed. Default: no.
+	For a list of relevant options see 'Options for UEFi executable' below.
 
 *-P, \--allpresets*::
 	Process all presets contained in '/etc/mkinitcpio.d'. See the '-p' option for
@@ -108,10 +86,6 @@ Options
 
 *-r, \--moduleroot* 'root'::
 	Specifies the root directory to find modules in, defaulting to '/'.
-
-*-R, \--osrelease* 'filename'::
-	Include a os-release file for the UEFI executable. Only used if '--uefi' is
-	specified. Default: '/etc/os-release' or '/usr/lib/os-release'.
 
 *-S, \--skiphooks* 'hooks'::
 	Skip 'hooks' when generating the image. Multiple hooks should be comma-separated.
@@ -136,6 +110,34 @@ Options
 
 *-z, \--compress* 'compress'::
 	Override the compression method with the 'compress' program.
+
+Options for UEFi executable
+---------------------------
+
+*--cmdline* 'config'::
+	Use kernel cmdline with UEFI executable. If none is specified it will
+	try find one of the files '/etc/kernel/cmdline', '/usr/share/kernel/cmdline'
+	or '/proc/cmdline'.
+
+*--splash* 'filename'::
+	UEFI executables can show a bitmap file on boot.
+
+*--uefistub* 'filename'::
+	UEFI stub image used for UEFI executable generation.
+	Default: Attempts to look for a systemd-boot or gummiboot
+	stub loader.
+
+*--kernelimage* 'filename'::
+	Include a kernel image for the UEFI executable. Default: one of
+	'/lib/modules/$KERNELVERSION/vmlinuz', '/boot/vmlinuz-$KERNELVERSION', or
+	'/boot/vmlinuz-linux'.
+
+*--microcode* 'filename'::
+	Include microcode into the UEFI executable. Default: no.
+
+*--osrelease* 'filename'::
+	Include a os-release file for the UEFI executable.
+	Default: '/etc/os-release' or '/usr/lib/os-release'.
 
 About Presets
 -------------
@@ -383,11 +385,11 @@ Examples
 	Create an initial ramdisk for the kernel at /boot/vmlinuz-linux. The
 	resulting image will be written to /boot/initramfs-linux.img.
 
-*mkinitcpio -g /boot/initramfs-linux.img -k /boot/vmlinuz-linux --uefi /efi/EFI/Linux/systemd-linux.efi*::
+*mkinitcpio -U /efi/EFI/Linux/systemd-linux.efi*::
 	Create an initial ramdisk for the kernel along with a UEFI executable.
 	The resuling executable will be written to /efi/EFI/Linux/systemd-linux.efi.
 
-*mkinitcpio -g /boot/initramfs-linux.img -k /boot/vmlinuz-linux -U /efi/EFI/Linux/systemd-linux.efi -M /boot/intel-ucode.img -l /usr/share/systemd/bootctl/splash-arch.bmp*::
+*mkinitcpio -U /efi/EFI/Linux/systemd-linux.efi --microcode /boot/intel-ucode.img --splash /usr/share/systemd/bootctl/splash-arch.bmp*::
 	Create an initial ramdisk for the kernel and an UEFI executable. This
 	also includes the Intel CPU microcode and a splash image which will be
 	used during boot.
--- a/man/mkinitcpio.8.txt
+++ b/man/mkinitcpio.8.txt
@@ -32,7 +32,7 @@ Options
 *-C, \--cmdline* 'config'::
 	Use kernel cmdline with UEFI executable. If none is specified it will
 	try find one of the files '/etc/kernel/cmdline', '/usr/share/kernel/cmdline'
-	or `/proc/cmdline`. Only used along with '--uefi'.
+	or '/proc/cmdline'. Only used along with '--uefi'.
 
 *-c, \--config* 'config'::
 	Use 'config' file to generate the ramdisk. Default: '/etc/mkinitcpio.conf'
--- a/man/mkinitcpio.8.txt
+++ b/man/mkinitcpio.8.txt
@@ -29,8 +29,13 @@ Options
 	after all other hooks from the config file. Multiple hooks should be
 	comma-separated. This option can be specified multiple times.
 
+*-C, \--cmdline* 'config'::
+	Use kernel cmdline with UEFI executable. If none is specified it will
+	try find one of the files '/etc/kernel/cmdline', '/usr/share/kernel/cmdline'
+	or `/proc/cmdline`. Only used along with '--uefi'.
+
 *-c, \--config* 'config'::
-	Use 'config' file to generate the ramdisk. Default: /etc/mkinitcpio.conf
+	Use 'config' file to generate the ramdisk. Default: '/etc/mkinitcpio.conf'
 
 *-d, \--generatedir* 'directory'::
 	Set 'directory' as the location where the initramfs is built. This might be
@@ -59,13 +64,38 @@ Options
 *-L, \--listhooks*::
 	List all available hooks.
 
+*-l, \--splash* 'filename'::
+	UEFI executables can show a bitmap file on boot.
+
+*-I, \--uefistub* 'filename'::
+	UEFI stub image used for UEFI executable generation. Only used if '--uefi'
+	is specified. Default: Attempts to look for a systemd-boot or gummiboot
+	stub loader.
+
+*-i, \--kernelimage* 'filename'::
+	Include a kernel image for the UEFI executable. This is only used if
+	'--uefi' is specified.  Default: one of
+	'/lib/modules/$KERNELVERSION/vmlinuz', '/boot/vmlinuz-$KERNELVERSION', or
+	'/boot/vmlinuz-linux'.
+
 *-M, \--automods*::
 	Display modules found via autodetection. mkinitcpio will automatically try to
 	determine which kernel modules are needed to start your computer. This option
 	lists which modules were detected.
 
+*-m, \--microcode* 'filename'::
+	Include microcode into the UEFI executable. This option is only used if
+	'--uefi' is specified. Default: no.
+
 *-n, \--nocolor*::
 	Disable color output.
+
+*-U, \--uefi* 'filename'::
+	Generate a UEFI executable as 'filename'. If a CPIO image is successfully
+	built, it will be used to generate a UEFI executable stub image for UEFI
+	booting. This combines the initramfs, the kernel, any specified microcode
+	and the kernel cmdline into one executable. This is useful for boot chain
+	integrity where the file is signed. Default: no.
 
 *-P, \--allpresets*::
 	Process all presets contained in '/etc/mkinitcpio.d'. See the '-p' option for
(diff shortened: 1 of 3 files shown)


mkinitcpio
```diff
@@ -251,6 +260,56 @@ build_image() {
...
+            cmdline="/etc/kernel/cmdline"
+        else
+            cmdline="/proc/cmdline"
+        fi
+    fi
+
+    if [[ -n "$splash" ]]; then
+        OBJCOPYARGS+=("--add-section .splash=$splash --change-section-vma .splash=0x40000")
+    fi
+    msg "Creating EFI image: %s" "$out"
```
Foxboron, line 297: "Creating UEFI executable" for consistency maybe?

mkinitcpio
```diff
@@ -251,6 +260,56 @@ build_image() {
...
+        for stub in {/usr,}/lib/systemd/boot/efi/linux{x64,ia32}.efi.stub {/usr,}/lib/gummiboot/linux{x64,ia32}.efi.stub; do
+            if [[ -f "$stub" ]]; then
+                uefistub="$stub"
+                break
+            fi
+        done
+    fi
+
+    if [[ -z "$kernelimg" ]]; then
+        for img in "/lib/modules/$KERNELIMAGE" "/boot/vmlinuz-$KERNELIMAGE" "/boot/vmlinuz-linuz"; do
```
Foxboron, line 278: This line is broken! Missing lookup with $KERNELVERSION D: and "vmlinuz-linuz"

mkinitcpio
```diff
@@ -251,6 +260,56 @@ build_image() {
...
+    if [[ -n "$splash" ]]; then
+        OBJCOPYARGS+=("--add-section .splash=$splash --change-section-vma .splash=0x40000")
+    fi
+    msg "Creating EFI image: %s" "$out"
+    objcopy \
+        --add-section .osrel="$osrelease" --change-section-vma .osrel=0x20000 \
+        --add-section .cmdline="$cmdline" --change-section-vma .cmdline=0x30000 \
+        --add-section .linux="$kernelimg" --change-section-vma .linux=0x2000000 \
+        --add-section .initrd=<(cat "$microcode" "$initramfs" ) --change-section-vma .initrd=0x3000000 \
+        ${OBJCOPYARGS[@]} "$uefistub" "$out"
```
grawity, line 303: Would it make sense to parse the $cmdline file, so that it could have embedded newlines and comments? The kernel command line can get a bit unwieldy, so my own scripts actually do something similar to:

```
--add-section .cmdline=<(grep '^[^#]' /etc/kernel/cmdline | tr -s '\n' ' ')
```

Admittedly this might result in a format that's incompatible with other existing implementations...

mkinitcpio
```diff
@@ -251,6 +260,56 @@ build_image() {
...
+    if [[ -z "$cmdline" ]]; then
+        if [[ -f "/etc/kernel/cmdline" ]]; then
+            cmdline="/etc/kernel/cmdline"
+        else
+            cmdline="/proc/cmdline"
+        fi
+    fi
+
+    if [[ -n "$splash" ]]; then
+        OBJCOPYARGS+=("--add-section .splash=$splash --change-section-vma .splash=0x40000")
```
grawity, line 295: This should probably be using the array like an array:

```
OBJCOPYARGS+=(--add-section .splash="$splash" --change-section-vma .splash=0x40000)
```

mkinitcpio
```diff
@@ -22,6 +22,8 @@ _d_presets=mkinitcpio.d
 # options and runtime data
 _optmoduleroot= _optgenimg=
 _optcompress= _opttargetdir=
+_optosrelease="/usr/lib/os-release"
```
grawity, line 25: This should first use /etc/os-release if that exists, and /usr/lib/os-release otherwise.

mkinitcpio
Foxboron: Argh, I forget the syntax for these things.

mkinitcpio.d/example.preset
```diff
@@ -8,6 +8,7 @@ PRESETS=('default' 'fallback')
 # as the path to an kernel image.
 ALL_kver='/boot/vmlinuz-linux'
 ALL_config='/etc/mkinitcpio.conf'
+ALL_microcode="/boot/intel-ucode.img"
```
grawity, line 11: In the real presets, will this somehow autodetect `amd-ucode.img`?

Foxboron: I'm not implementing cross-distro autodetection of microcode updates.

We can make this and array and cat together every file it finds. But I think it's better to let the user decide.

grawity: Then maybe it's better to make it an array, e.g. `(/boot/*-ucode.img)` by default. So far users haven't had to edit presets by default, I'm not really sure if it's a good idea to add yet another place that must be changed during installation.

Foxboron: I don't think having UEFI executable generation on by *default* is the goal. This is going to be optional. Manual configuration for that purpose is fine.

coderobe: But also, why make it harder when the array option doesn't seem like a significant amount of work to include?
If it's purely an effort kinda deal, I'd be up for fixing this in a follow-up patch.

Foxboron: That wasn't an argument against using arrays. It was more about supplying default configurations or not.

mkinitcpio
```diff
@@ -251,6 +260,56 @@ build_image() {
...
+                break
+            fi
+        done
+    fi
+
+    if [[ -z "$cmdline" ]]; then
+        if [[ -f "/etc/kernel/cmdline" ]]; then
+            cmdline="/etc/kernel/cmdline"
+        else
+            cmdline="/proc/cmdline"
```
coderobe, line 290: Perhaps it would make sense to spit out a warning/info blip when the current `cmdline` is reused here instead of the config in /etc.
```suggestion
            msg2 "Note: /etc/kernel/cmdline does not exist and --cmdline is unset!"
            cmdline="/proc/cmdline"
            msg2 "Reusing current kernel cmdline from $cmdline"
```

mkinitcpio
```diff
@@ -400,6 +474,13 @@ while :; do
         -n|--nocolor)
             _optcolor=0
             ;;
+        -U|--uefi)
+            shift
+            [[ -d $1 ]] && die "Invalid image path -- must not be a directory"
+            if ! _optuefi=$(readlink -f "$1") || [[ ! -e ${_optgenimg%/*} ]]; then
+                die "Unable to write to path: \`%s'" "$1"
```
coderobe, line 481: Is there a reason the first quote here is a backtick?

mkinitcpio.d/example.preset
```diff
@@ -17,9 +18,11 @@ ALL_config='/etc/mkinitcpio.conf'
 #default_kver="3.0-ARCH"
 #default_config="/etc/mkinitcpio.conf"
 default_image="/tmp/initramfs-linux.img"
+default_efi_image="/efi/EFI/Linux/linux.efi"
```
coderobe, line 21: This should perhaps be namespaced to avoid messing with foreign linuxes under EFI/

mkinitcpio.d/example.preset
```diff
@@ -17,9 +18,11 @@ ALL_config='/etc/mkinitcpio.conf'
...
 default_options=""
 
 #fallback_kver="3.0-ARCH"
 #fallback_config="/etc/mkinitcpio.conf"
 fallback_image="/tmp/initramfs-linux-fallback.img"
+fallback_efi_image="/efi/EFI/Linux/linux-fallback.efi"
```
coderobe, line 27: Same as above

mkinitcpio
Foxboron: I realized I should add a `msg2` for all of the stuff we detect if it's not defined. Same applies for `$uefistub` above.

mkinitcpio
```diff
@@ -251,6 +260,56 @@ build_image() {
...
+    local out=$1 microcode=$2 initramfs=$3 cmdline=$4 osrelease=$5 splash=$6 kernelimg=$7 uefistub=$8 errmsg=
+    OBJCOPYARGS=()
+
+    if [[ -z "$uefistub" ]]; then
+        for stub in {/usr,}/lib/systemd/boot/efi/linux{x64,ia32}.efi.stub {/usr,}/lib/gummiboot/linux{x64,ia32}.efi.stub; do
+            if [[ -f "$stub" ]]; then
+                uefistub="$stub"
+                break
+            fi
+        done
```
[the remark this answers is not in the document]
Foxboron: The stat line feels to clever. I need to think twice to realize why it's written like that and feels more like golfing then being clean and easy to understand.

mkinitcpio
Foxboron: It's inherited from some other code. I have no clue.

mkinitcpio.d/example.preset
Foxboron: Namespaces how? `/efi/EFI/Arch/linux.efi` or `/efi/EFI/Linux/arch-linux.efi`?

mkinitcpio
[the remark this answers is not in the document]
coderobe: In that case, opinions on just shortening the parameter expansion?
```suggestion
        for stub in {/usr,}/lib/{systemd/boot/efi,gummiboot}/linux{x64,ia32}.efi.stub; do
            if [[ -f "$stub" ]]; then
                uefistub="$stub"
                break
            fi
        done
```

mkinitcpio.d/example.preset
coderobe: I haven't come up with any implementation details; The latter sounds fine to me, if the former is not an option. I just think that we should have _some_ kind of namespacing, else we might clobber existing stuff from elsewhere, the chances of which I'd like to minimize.

mkinitcpio
Foxboron: This is the upstream code utilizing "/etc/kernel/cmdline" (i'm realizing I need to have better fallbacks to /usr/lib).

https://github.com/systemd/systemd/blob/183d5168cf1d2bfbd3049e8592b4ed4c87c94936/src/kernel-install/90-loaderentry.install#L65-L78

I think this would be a bad(?) idea and you are better off supplying this as your own option?

mkinitcpio
Maryse47: Some tools do backups of `/lib/modules/$KERNELIMAGE` in roder to not lose modules of running kernel after live update so maybe it's better to skip that path and use only `/boot`

grawity: > Some tools do backups of `/lib/modules/$KERNELIMAGE` in order to not lose modules of running kernel after live update so maybe it's better to skip that path and use only `/boot`

That shouldn't be a problem; mkinitcpio already knows which kernel version it's being invoked for (I mean, a _large_ part of its job is to copy the correct modules from /lib/modules) so it'll just take the kernel image from the same directory, I assume.

mkinitcpio.d/example.preset
Foxboron: This was painfull, but implemented!

mkinitcpio
Foxboron: I don't consider that a problem. We should know which kernel we are building the initramfs for. If someone points the wrong kernel at it without any modules in `/lib/modules` then it's a user error. We issue an error then anyway.

Maryse47: I meant the opposite situation - the modules exist but kernel at `/boot` doesn't however if the version is known beforehand then it should be ok, thx.

Foxboron: If the modules directory exists and `/boot` doesn't have a kernel, then you should still have `/lib/modules/$KERNELVERSION/vmlinuz`?

Maryse47: Yes but building image out of it is never good idea as it may lead to issues like [this](https://github.com/andreyv/sbupdate/issues/38) one. This is why I wanted to confirm mkinitcpio won't be affected by same problem.

Foxboron: That issue is rooted in `sbupdate` being unaware of the correct kernel version and guessing from the modules directory. That is a mistake which can't happen in `mkinitcpio`.

Foxboron:
Added a manpage. Missing some '/usr/lib/os-release' fallback but should be'ish fine I think. @grazzolini I think a general ack/nack or changes you want to see would be useful at this point.

mkinitcpio
```diff
@@ -251,6 +260,79 @@ build_image() {
...
 }
 
+build_uefi(){
+    local out=$1 initramfs=$2 cmdline=$3 osrelease=$4 splash=$5 kernelimg=$6 uefistub=$7 microcode=(${@:7}) errmsg=
+    OBJCOPYARGS=()
+
+    msg "Creating UEFI executable: %s" "$out"
+
+    if [[ -z "$uefistub" ]]; then
+        for stub in {/usr,}/lib/{systemd/boot/efi,gummiboot}/linux{x64,ia32}.efi.stub; do
```
gdamjan, line 270: the list this generates is:
```
/usr/lib/systemd/boot/efi/linuxx64.efi.stub
/usr/lib/systemd/boot/efi/linuxia32.efi.stub
/usr/lib/gummiboot/linuxx64.efi.stub
/usr/lib/gummiboot/linuxia32.efi.stub
/lib/systemd/boot/efi/linuxx64.efi.stub
/lib/systemd/boot/efi/linuxia32.efi.stub
/lib/gummiboot/linuxx64.efi.stub
/lib/gummiboot/linuxia32.efi.stub
```
just to make sure that's the intended behavior

mkinitcpio
gdamjan: the "upstream" `read -r -d '' -a BOOT_OPTIONS` does handle `\n` as a space

I'd personally prefer grawity's version where we can also add comments, but maybe for a future RFE?

gdamjan:
It  would be useful to have an Example in the man page, for manual non-preset invocation,

mkinitcpio
Foxboron: Yep, that seems correct. It should support !usrmerge, and any distros that doesn't ship systemd.

Foxboron:
Manpages has been fixed up. I intend to try expand the preset documentation in another PR. It should preferably explain how it works instead of you having to infer it through examples.

mkinitcpio
Foxboron: I went with grawity's version and waiting on feedback from grazzolini if it's okay or not :)

mkinitcpio
Foxboron: I forgot it, but fixed now.

Foxboron:
I think everything should be good. I have built a package people can test if they want to.

https://pkgbuild.com/~foxboron/repos/loose-pkgs/

https://github.com/Foxboron/PKGBUILDS/tree/master/mkinitcpio

Foxboron:
I realized #8 is a thing. I hijacked `-R` for `--osrelease`. Does anyone have a better flag suggestion :D?

klausenbusk:
> I think everything should be good. I have built a package people can test if they want to.

I tested this and it works like a dream, good work! :)

gdamjan:
> I realized #8 is a thing. I hijacked `-R` for `--osrelease`. Does anyone have a better flag suggestion :D?

-O ?

Foxboron:
I have no clue how I missed the availability of `-O`...

Foxboron:
I'd be happy to drop all the short options (except for `-U`) if @grazzolini thinks it's fine. I have mostly just mirrored the existing stuff which all have both.

gdamjan:
First of all, great feature. Just a simple suggestion. 

Currently the command `mkinitcpio --uefi /boot/EFI/Linux/xyz.efi` finishes with `Dry run complete, use -g IMAGE to generate a real image`.

In my opinion, it makes sense that the command already had an image specified , and it should just make the initramfs in /tmp/<tmpdir>/ and then create the final uefi image it was asked for.

Foxboron:
Frankly I'd prefer if `mkinitcpio -p linux --uefi /boot/EFI/Linux/xyz.efi` would be possible which is what https://github.com/archlinux/mkinitcpio/pull/56 introduces.

I agree that `mkinitcpio --uefi` should do the correct thing but I'm a bit unsure if this is a super ugly hack or not :)

```diff
diff --git a/mkinitcpio b/mkinitcpio
index 0111f80..eda4dca 100755
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -599,6 +599,14 @@ if [[ -n $_d_flag_hooks && -n $_d_flag_install ]]; then
     _d_install=${_d_flag_install%:}
 fi
 
+
+# If we specified --uefi but no -g we want to create a temporary initramfs which will be used with the efi executable.
+if [[ $_optuefi && $_optgenimg == "" ]]; then
+    tmpfile=$(mktemp -t mkinitcpio.XXXXXX)
+    trap "rm $tmpfile" EXIT
+    _optgenimg="$tmpfile"
+fi
+
 # insist that /proc and /dev be mounted (important for chroots)
 # NOTE: avoid using mountpoint for this -- look for the paths that we actually
 # use in mkinitcpio. Avoids issues like FS#26344.
```

man/mkinitcpio.8.txt
```diff
@@ -59,14 +64,39 @@ Options
...
+
+*-I, \--uefistub* 'filename'::
+	UEFI stub image used for UEFI executable generation. Only used if '--uefi'
+	is specified. Default: Attempts to look for a systemd-boot or gummiboot
+	stub loader.
+
+*-i, \--kernelimage* 'filename'::
+	Include a kernel image for the UEFI executable. This is only used if
+	'--uefi' is specified.  Default: one of
+	'/lib/modules/$KERNELVERSION/vmlinuz', '/boot/vmlinuz-$KERNELVERSION', or
```
grazzolini, line 78: Can't this be derived from the preset, if not specified? If I'm creating an UEFI executable for -lts, for example, the preset already lists what is the kernel image and the path.

Foxboron: It is derived though, but we can't be sure where the kernel image is in all cases which is why we also expose the ability to override whatever mkinitcpio tries to figure out. `--kernel` which you are probably thinking about is for the kernel version, not the file itself.

Foxboron:
I have removed the short options, fixed up the manpage+help text. I also implemented the `-U` workaround with a temporary initramfs as suggested.

grazzolini (approved):

Foxboron:
@grazzolini I'm not quite sure why you decided to merge this with 11 fixup commits I intended to squash before finalizing. The idea was to make it easier to review incremental changes.

yan12125:
@Foxboron Huge thanks for this feature! Here is a missing fix - the short option `-m` for `--microcode` is dropped, while the mkinitcpio script still uses `-m` for microcode imags parsed from preset files [1]

[1] https://github.com/archlinux/mkinitcpio/blob/f40bb42577284083adaf1edca2f05300badb78b2/mkinitcpio#L415

Foxboron:
@yan12125 ohnoo! I'll send a fix :)

grazzolini:
> @grazzolini I'm not quite sure why you decided to merge this with 11 fixup commits I intended to squash before finalizing. The idea was to make it easier to review incremental changes.

I don't mind the fixups, don't worry. I plan on making a new release soon.

mkinitcpio
```diff
@@ -345,10 +453,11 @@ preload_builtin_modules() {
 trap 'cleanup 130' INT
 trap 'cleanup 143' TERM
 
-_opt_short='A:c:D:g:H:hk:nLMPp:r:S:sd:t:Vvz:'
+_opt_short='A:c:D:g:H:hk:nLMPp:r:S:sd:t:U:Vvz:'
 _opt_long=('add:' 'addhooks:' 'config:' 'generate:' 'hookdir': 'hookhelp:' 'help'
           'kernel:' 'listhooks' 'automods' 'moduleroot:' 'nocolor' 'allpresets'
-          'preset:' 'skiphooks:' 'save' 'generatedir:' 'builddir:' 'version' 'verbose' 'compress:')
+          'preset:' 'skiphooks:' 'save' 'generatedir:' 'builddir:' 'version' 'verbose' 'compress:'
+          'uefi:' 'microcode:' 'splash:' 'kernelimage:' 'uefistub:')
```
jirutka, line 460: You forgot to add option `cmdline` here.

mkinitcpio
```diff
@@ -251,6 +262,92 @@ build_image() {
...
+        elif [[ -f "/usr/lib/kernel/cmdline" ]]; then
+            cmdline="/usr/lib/kernel/cmdline"
+        else
+            warning "Note: /etc/kernel/cmdline does not exist and --cmdline is unset!"
+            cmdline="/proc/cmdline"
+            warning "Reusing current kernel cmdline from $cmdline"
+        fi
+        msg2 "Using cmdline file %s" "$cmdline"
+    fi
+    if [[ ! -f "$cmdline" ]]; then
```
jirutka, line 310: Either this condition is wrong, or the description of the `--cmdline` option is misleading.

> Set kernel line (default content: /etc/kernel/cmdline, /proc/cmdline)

I thought that I can just directly pass the kernel line, but it actually expects a file.

Foxboron: It's expecting a file so it should *probably* mention it, but you could just pass `--cmdline < <(echo "rw root=/dev/mapper/root")` or something similar i reckon.

jirutka: Not probably, but definitely; it’s very unclear now. ;)

Process substitution (`<( cmd )`) is not defined by POSIX command language, nor supported by majority of the shells, just by Bash and ZSH, so it’s not portable. Sadly, mkinitcpio contains a lot of bashisms, but one must have Bash installed to use it, but that doesn’t mean it will be always called from Bash.

Now I have to do the following:

```sh
cmdline_file=$(mktemp)
printf %s "$cmdline" > "$cmdline_file"

mkinitcpio --cmdline "$cmdline_file" ...

rm -f "$cmdline_file"
```

…which is suboptimal.

Allowing to just pass cmdline as a value would be IMHO more convenient in general.

gdamjan: > Allowing to just pass cmdline as a value would be IMHO more convenient in general. Just a suggestion.

I guess that depends what you do with mkinitcpio.

I've almost never wanted to specify it as an argument, as opposed to having it in /etc/kernel/cmdline

Foxboron: Supporting this as a string when dealing with the preset files and stuff would just be tedious and a lot harder to implement correctly. I'm happy with this being a file and this is what other initramfs generators following `kernel-install` and `dracut` does as well.

T35R6braPwgDJKq:
man page suggests that all `-U` options fall back to default, yet a simple `mkinitcpio -U` (expecting to fallback to said defaults) fails with: `mkinitcpio: option '-U' requires an argument`. That seems incoherent or at least inaccurate

PR #54, opened 2021-04-07
functions: move add_udev_rule() from systemd hook

eworm-de:
This function can be used for setups with systemd in initramfs (hook
`systemd`) or without (hook `udev`). To avoid code duplication move
the code here.

### DIFF
--- a/functions
+++ b/functions
@@ -625,6 +625,46 @@ add_binary() {
     return 0
 }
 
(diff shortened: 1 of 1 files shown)


eworm-de:
Using this code in systemd 248-4... Should be reasonable stable.

grazzolini (approved):

PR #72, opened 2021-11-21
fsck root device: respect provided rootfstype

nightfog-git:
By default, `fsck` determines filesystem type with a lookup of `/etc/fstab`. As for now it's empty in initramfs. So it defaults to ext2. That could (and actually do) damage the filesystem if it's not of an ext-family. The `-a` option is propagated to fsck.ext2 and it make it worse. 
The solution is `fsck -t` option.  It will run the correct `fsck.$rootfstype` and not making harm with wrong one.


grazzolini:
This is already correctly handled, see the discussion on [FS#72806](https://bugs.archlinux.org/task/72806).

PR #75, opened 2021-12-09
functions: allow tailing `+` for custom kernel build upon git trees

adam900710:
For kernels build upon git trees, any non-tag commit will append a '+'
char to kernel version.

This means the current fallback method will cut off the tailing '+' and
cause mkinitcpio failure for suck cases.

Signed-off-by: Qu Wenruo <wqu@suse.com>

### DIFF
--- a/functions
+++ b/functions
@@ -158,7 +158,7 @@ kver_generic() {
     # https://elixir.bootlin.com/linux/v5.7.2/source/init/version.c#L46
     local kver=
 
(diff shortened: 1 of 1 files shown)


nl6720:
I'd need an example version string to review this.

adam900710:
The example string would be `6.1.0-rc1-custom+`

nl6720 (approved):

Issue #12, opened 2019-11-17
Support signing the kernel and/or initramfs/microcode (secure boot)


klausenbusk:
If people want to use Secure Boot with there own keys the kernel (and preferable also the initramfs, cmdline and microcode) needs to be signed.

The can be done with a custom hook like this:
```
[Trigger]
Operation = Install
Operation = Upgrade
Type = Package
Target = linux

[Action]
Description = Signing Kernel for SecureBoot
When = PostTransaction
Exec = /usr/bin/sbsign --key /root/db.key --cert /root/db.crt /boot/vmlinuz-linux
```
Would it make sense to support this out-of-the-box? Something like this:
```patch
diff --git a/libalpm/scripts/mkinitcpio-install b/libalpm/scripts/mkinitcpio-install
index d2e5041..213c174 100644
--- a/libalpm/scripts/mkinitcpio-install
+++ b/libalpm/scripts/mkinitcpio-install
@@ -30,6 +30,11 @@ while read -r line; do
     # always install the kernel
     install -Dm644 "${line}" "/boot/vmlinuz-${pkgbase}"
 
+    # sign the kernel
+    if [[ -n "$secureboot_db_key" && -n "$secureboot_db_key" ]]; then
+      sbsign --key "${secureboot_db_key}" --cert "${secureboot_db_key}" "/boot/vmlinuz-${pkgbase}"
+    fi
+
     # compound args for each kernel
     args+=(-p "${pkgbase}")
 done
diff --git a/mkinitcpio.d/hook.preset b/mkinitcpio.d/hook.preset
index 66709a8..d04d035 100644
--- a/mkinitcpio.d/hook.preset
+++ b/mkinitcpio.d/hook.preset
@@ -12,3 +12,6 @@ default_image="/boot/initramfs-%PKGBASE%.img"
 #fallback_config="/etc/mkinitcpio.conf"
 fallback_image="/boot/initramfs-%PKGBASE%-fallback.img"
 fallback_options="-S autodetect"
+
+#secureboot_db_key=""
+#secureboot_db_cert=""
```

[Ideally](https://mjg59.dreamwidth.org/48897.html) we should also sign the initramfs (+ microcode) and cmdline, which can be done by using the `systemd` efi stub, but I'm not sure how to do that cleanly:
```shell
cat /boot/intel-ucode.img /boot/initramfs-linux.img > initramfs.img

objcopy \
[...]

PR #79, opened 2021-12-20
Add support for UEFI executable siging

gekmihesg:
Allows setting `ALL_efi_cert`/`<preset>_efi_cert` and `ALL_efi_key`/`<preset>_efi_key` in presets to automatically sign generated EFI executables.

At the moment there doesn't seem to be a clean way to do this otherwise. Using a pacman hook only partially works and would probably leave the user with a unbootable system when running `mkinitcpio` manually.


Foxboron:
I'm of the opinion that this doesn't belong in an initramfs generator and should be implemented as a hook, or have mkinitcpio support post-generation commands.

Depending on sbsign and plain-text certificates shouldn't be a thing.

Obviously I'm biased towards a solution like https://github.com/Foxboron/sbctl

mkinitcpio
```diff
@@ -277,7 +280,7 @@ build_uefi(){
                 break
             fi
         done
-    elif [[ ! -f "$uefisub" ]]; then
+    elif [[ ! -f "$uefistub" ]]; then
```
Foxboron, line 283: This is a separate bugfix and should not be hidden in this commit :)

Foxboron (changes requested):

mkinitcpio
gekmihesg: Sure, moved it to #80. Was broken in some other places anyway.

Foxboron: Thanks! You didn't drop the commit to the master branch though :)

gekmihesg: Should disappear from this PR once #80 gets merged I think. At least it's the same commit in both PRs. If not, I'll fix that. At the moment I get the feeling, that this PR won't get through anyway ;-)

Foxboron: Please don't let my opinion discourage you :) @grazzolini is still the maintainer and if he want's this I'll help ensure it's as correct as it can be despite my objections.

gekmihesg:
I read through #12 and I can somewhat agree. Although, since mkinitcpio provides a feature to create a bundled efi binary, there should be some way to sign it automatically. I can't really decide on how I feel about the bundling process being part of mkinitcpio but not the signing part. But since it is there, I think there should be some way to sign the bundle. At least making this possible with a generic post-generation hook would be an option as well (and also way more flexible). But then maybe the whole bundling process should have been located there as well, since to me this feels as related to generating the initramfs itself as the signing does ;-)

Foxboron:
Having UEFI stubs is not *only* for Secure Boot though. They simplify the boot setup when used in conjunction with `sd-boot`, so I don't buy the argument which boils down to "signing shouldn't be inside mkinitcpio, therefore unified kernel images shouldn't either".

Having the current unified kernel stuff in it's own thing would be more complicated and can't piggyback that easily on the existing configuration and preset stuff that `mkinitcpio` has. However for Secure Boot there isn't even any reason to support presets either. You are most likely going to be using one set of keys.

The main argument I have against this is that it enforces a singular way of having keys generated by `openssl` without support for the different openssl engines. You can in theory have keys using PKCS11 and PIV inside the TPM or a yubikey. But none of these implementations (looking at this and dracut) exposes it. I think that is a bad option. Supporting all these engines in a good way would also not be trivial I think.

Thus I'd much rather see some post-exec hooks *inside* mkinitcpio where the sbsign doesn't need to be hardcoded, but provided the proper mechanics without having to hardcode anything. It would also accomplish the same goal: "I don't want `mkinicpio -P` to produce unsigned kernel images".

grazzolini:
I think having supports for post generation hooks would be way better than hardcoding this in one manner. Even if we add knobs to support agnostic stuff, it is hard to test this.

Having support for this adds tons of possibilities, for example, a hook that gets shipped with sbsigntools or sbctl, each doing their own thing.

PR #81, opened 2021-12-20
Add support for post generation hooks

gekmihesg:
Alternative approach to #79

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -14,8 +14,10 @@ _f_functions=functions
 _f_config=mkinitcpio.conf
 _d_hooks="$PWD/hooks:/usr/lib/initcpio/hooks:/lib/initcpio/hooks"
 _d_install="$PWD/install:/usr/lib/initcpio/install:/lib/initcpio/install"
+_d_post="$PWD/post:/usr/lib/initcpio/post:/lib/initcpio/post"
 _d_flag_hooks=
 _d_flag_install=
+_d_flag_post=
 _d_firmware=({/usr,}/lib/firmware/updates {/usr,}/lib/firmware)
 _d_presets=mkinitcpio.d
 
@@ -26,7 +28,7 @@ _optosrelease=
 _optuefi= _optmicrocode=() _optcmdline= _optsplash= _optkernelimage= _optuefistub=
 _optshowautomods=0 _optsavetree=0 _optshowmods=0
 _optquiet=1 _optcolor=1
-_optskiphooks=() _optaddhooks=() _hooks=()  _optpreset=()
+_optskiphooks=() _optaddhooks=() _hooks=()  _optpreset=() _generated=()
 declare -A _runhooks _addedmodules _modpaths _autodetect_cache
 
 # export a sane PATH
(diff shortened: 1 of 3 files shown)


Project0:
I am currently workaround with a pacman hook, but this is not always catching the rebuild of a image. This would solve the issue :+1:

Molytho (approved):

Molytho:
Maybe there could be a way (e.g when script exits with 255) to notify mkinitcpio about a fatal error so that for example when signing fails the build process fails with a "fatal" instead of only a "warning"

gekmihesg:
Should we just assume that all errors are fatal? A warning could also be printed by the hook executable itself.
Either way, what should happen on a fatal error? Stop processing other hooks and make mkinitpio exit 1?

Molytho:
I'm fine with just always assuming a fatal error.
Exiting with 1 and stop processing is what I thought of.

I just think that the user should get a proper feedback for error when a hook is crucial for the specific setup (like secureboot signing).
Also (somewhere in the future) there might be hooks that depend on other hooks to be run in advance and just continuing after a hook failed might break there assumptions.

gekmihesg:
Agreed, fixed

mkinitcpio
```diff
@@ -458,6 +460,37 @@ preload_builtin_modules() {
...
+    local args=("$@")
+    local hook status
+    local -i count=0
+    local -a paths seen
+
+    IFS=: read -ra paths <<<"$_d_post"
+
+    for path in "${paths[@]}"; do
+        for hook in "$path"/*; do
+            [[ -e $hook ]] && ! in_array "${hook##*/}" "${seen[@]}" || continue
```
IsaacG, line 473:
```suggestion
            if [[ -e "$hook" ]] && ! in_array "${hook##*/}" "${seen[@]}"; then
                continue
            fi
```

mkinitcpio
```diff
@@ -458,6 +460,37 @@ preload_builtin_modules() {
...
+            seen+=("${hook##*/}")
+
+            [[ -x $hook ]] || continue
```
IsaacG, line 476:
```suggestion
            [[ -x "$hook" ]] || continue
```

mkinitcpio
```diff
@@ -458,6 +460,37 @@ preload_builtin_modules() {
...
+            status=$?
+
+            if (( status )); then
+                error "'%s' failed with exit code %d" "$hook" "$status"
+                return 1
+            fi
+        done
+    done
+
+    ! (( count )) || msg "Post processing done"
```
IsaacG, line 491:
```suggestion
    (( count )) && msg "Post processing done"
```

mkinitcpio
```diff
@@ -608,9 +642,10 @@ if [[ -t 1 ]] && (( _optcolor )); then
     try_enable_color
 fi
 
-if [[ -n $_d_flag_hooks && -n $_d_flag_install ]]; then
+if [[ -n $_d_flag_hooks && -n $_d_flag_install && -n $_d_flag_post ]]; then
```
IsaacG, line 645:
```suggestion
if [[ -n "$_d_flag_hooks" && -n "$_d_flag_install" && -n "$_d_flag_post" ]]; then
```

mkinitcpio
```diff
@@ -714,15 +749,21 @@ ldconfig -r "$BUILDROOT" &>/dev/null
 umask 077
 
 if [[ $_optgenimg ]]; then
-    build_image "$_optgenimg" "$_optcompress"
+    build_image "$_optgenimg" "$_optcompress" &&
+        _generated+=("$_optgenimg")
```
IsaacG, line 753:
```suggestion
    build_image "$_optgenimg" "$_optcompress" && _generated+=("$_optgenimg")
```

klausenbusk:
Moved to to: https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/merge_requests/81.

PR #82, opened 2021-12-21
mkinitcpio: Support /etc/kernel/cmdline.d

Foxboron:

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -264,7 +264,7 @@ build_image() {
 }
 
 build_uefi(){
-    local out=$1 initramfs=$2 cmdline=$3 osrelease=$4 splash=$5 kernelimg=$6 uefistub=$7 microcode=(${@:7}) errmsg=
+    local out=$1 initramfs=$2 cmdline=($3) osrelease=$4 splash=$5 kernelimg=$6 uefistub=$7 microcode=(${@:7}) errmsg=
     OBJCOPYARGS=()
 
     msg "Creating UEFI executable: %s" "$out"
@@ -298,15 +298,18 @@ build_uefi(){
 
     if [[ -z "$cmdline" ]]; then
         if [[ -f "/etc/kernel/cmdline" ]]; then
-            cmdline="/etc/kernel/cmdline"
+            cmdline+=("/etc/kernel/cmdline")
         elif [[ -f "/usr/lib/kernel/cmdline" ]]; then
-            cmdline="/usr/lib/kernel/cmdline"
+            cmdline+=("/usr/lib/kernel/cmdline")
         else
             warning "Note: /etc/kernel/cmdline does not exist and --cmdline is unset!"
-            cmdline="/proc/cmdline"
+            cmdline+=("/proc/cmdline")
             warning "Reusing current kernel cmdline from $cmdline"
         fi
-        msg2 "Using cmdline file: %s" "$cmdline"
+        if [[ -d "/etc/kernel/cmdline.d" ]]; then
+            cmdline+=(/etc/kernel/cmdline.d/*.conf)
+        fi
+        msg2 "Using cmdline file: %s" "${cmdline[@]}"
     fi
     if [[ ! -f "$cmdline" ]]; then
         error "Kernel cmdline file '%s' not found" "$cmdline"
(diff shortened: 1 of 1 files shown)


Project0:
That is exactly what i was looking for :D

WhyNotHugo:
Oh, nice to see this happening. Maybe `man` page can mention this too? The docs for `--cmdline` describe existing fallbacks when no value is specified via the command line.

mkinitcpio
```diff
@@ -298,15 +298,18 @@ build_uefi(){
...
-            cmdline="/usr/lib/kernel/cmdline"
+            cmdline+=("/usr/lib/kernel/cmdline")
         else
             warning "Note: /etc/kernel/cmdline does not exist and --cmdline is unset!"
-            cmdline="/proc/cmdline"
+            cmdline+=("/proc/cmdline")
             warning "Reusing current kernel cmdline from $cmdline"
         fi
-        msg2 "Using cmdline file: %s" "$cmdline"
+        if [[ -d "/etc/kernel/cmdline.d" ]]; then
```
WhyNotHugo, line 309: The previous `else` will match if neither `/etc/kernel/cmdline` nor `/usr/lib/kernel/cmdline` exist. This prints that `/proc/cmdline` will be use and puts if first on the list, but then the files in `cmdline.d` get read.

It would seem the values in `cmdline.d` never get effectively used.

Foxboron:
I'm contemplating redoing some of this to better synergize with dracut which uses `/etc/cmdline.d` and not `/etc/kernel/cmdline{,.d}`

mkinitcpio
Foxboron: The else clause doesn't have anything to do with the if-section for `cmdline.d` so I'm confused why you think it won't be used? I personally use this in one of my laptops currently.

Project0:
While i was building something similar to workaround the lack of this, i realized i need different kernel options per build preset.
Would this something which would make sense to include?

Looks like this in my case:
```ls
# default for all
/etc/kernel/cmdline.preset.d/:
00-project0-base-config-ro  20-project0-desktop-quiet  40-project0-cryptsetup

# additions per preset
/etc/kernel/cmdline.preset.d/default:
10-project0-base-config-nobgrt  50-local-fast-again

/etc/kernel/cmdline.preset.d/fallback:

/etc/kernel/cmdline.preset.d/terminal:
10-project0-base-config-sd-target
```

Foxboron:
I'm against overcomplicating it. I'd much rather do some changes to `--cmdline` so you can do


<pre>
PRESETS=(default fallback terminal)

default_options="--cmdline /etc/kernel/cmdline.preset.d/project0-base-config-ro --cmdline /etc/kernel/cmdline.preset.d/50-local-fast-again"
fallback_options=""
terminal_options="--cmdline /etc/kernel/cmdline.preset.d/10-project0-base-config-sd-target
<pre>

Or something along those lines.

Maryse47:
> I'm contemplating redoing some of this to better synergize with dracut which uses `/etc/cmdline.d` and not `/etc/kernel/cmdline{,.d}`

/etc/kernel/cmdline is used by [kernel-install](https://www.freedesktop.org/software/systemd/man/kernel-install.html) so maybe it should be kept.

Foxboron:
Hmm, should it be
* `/etc/kernel/cmdline`
* `/etc/cmdline.d`
Should we retain the kernel prefix?

mkinitcpio
```diff
@@ -298,15 +298,18 @@ build_uefi(){
...
+            cmdline+=(/etc/kernel/cmdline.d/*.conf)
+        fi
+        msg2 "Using cmdline file: %s" "${cmdline[@]}"
```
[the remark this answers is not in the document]
nl6720: No need to bring the variable inside the string.

```suggestion
        msg2 'Using cmdline file: %s' "${cmdline[*]}"
```

Foxboron:
Going with the suggestion from @Maryse47  and postponing it until v33 :)

klausenbusk:
Moved to to: https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/merge_requests/82.

PR #83, opened 2021-12-21
mkinitcpio: add VERSION_ID to EFISTUB

Foxboron:
This patch adds VERSION_ID to the built images. This helps distinguish
between several kernel versions instead of having it fallback to showing
file names. The sd-boot menu shows the new entires with the given kernel
ID:

Old menu:
    * Arch Linux (arch-systemd.efi)
    * Arch Linux (test-efi.efi)

New menu:
    * Arch Linux (rolling)
    * Arch Linux (5.15.5-arch1-1)

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -341,7 +341,7 @@ build_uefi(){
     done
 
     objcopy \
-        --add-section .osrel="$osrelease" --change-section-vma .osrel=0x20000 \
+        --add-section .osrel=<(cat "$osrelease" && echo "VERSION_ID=$KERNELVERSION") --change-section-vma .osrel=0x20000 \
         --add-section .cmdline=<(grep '^[^#]' "$cmdline" | tr -s '\n' ' ') --change-section-vma .cmdline=0x30000 \
         --add-section .linux="$kernelimg" --change-section-vma .linux=0x2000000 \
         --add-section .initrd=<(cat ${microcode[@]} "$initramfs") --change-section-vma .initrd=0x3000000 \


Foxboron:
@kyrias and I discussed on IRC if there was something more suitable.. I lost the log and forgot their suggestion. I think it was `BUILD_ID`?

Priority can be found here: https://github.com/systemd/systemd/blob/main/src/fundamental/bootspec-fundamental.c#L44

gekmihesg:
With the default config, this wouldn't really help because it generates a `default` and a `fallback` version of the image, both with the same `KERNELVERSION`.
Maybe that could be passed as a param from the `process_preset`? Something like
```sh
imageversion="$preset-$p"
preset_cmd+=(--imageversion "${imageversion%-default}")
```
```sh
... <(cat "$osrelease" && echo "IMAGE_VERSION=${_optimageversion:-$KERNELVERSION"})
```

Foxboron:
Ah, I always forget the fallback, so having the preset name included in the kernel version would probably be a good idea.

gekmihesg:
Personally, I would not include the kernel version but just use it as a fallback when nothing else is specified. It looks a little messy otherwise. For me that would be something like this:
```
Arch Linux (5.15.10-arch1-1-linux-fallback)
Arch Linux (5.10.87-1-lts-linux-lts-fallback)
Arch Linux (5.10.87-1-lts-linux-lts)
Arch Linux (5.15.10-arch1-1-linux)
```
Without:
```
Arch Linux (linux-fallback)
Arch Linux (linux-lts-fallback)
Arch Linux (linux-lts)
Arch Linux (linux)
```

Foxboron:
Hmm, but what in the cases where people create i stubs for the sake of having rollback support? Preferably we should be able to distinguish between the presets and the kernel versions as well. Should we alter the name to include the kernel version?

Example
```
Arch Linux 5.15.10-arch1-1 (linux)
```

mkinitcpio
```diff
@@ -341,7 +341,7 @@ build_uefi(){
     done
 
     objcopy \
-        --add-section .osrel="$osrelease" --change-section-vma .osrel=0x20000 \
+        --add-section .osrel=<(cat "$osrelease" && echo "VERSION_ID=$KERNELVERSION") --change-section-vma .osrel=0x20000 \
```
IsaacG, line 344:
```suggestion
        --add-section .osrel=<(cat "$osrelease"; echo "VERSION_ID=$KERNELVERSION") --change-section-vma .osrel=0x20000 \
```

klausenbusk:
Migrated to to: https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/merge_requests/83.

PR #86, opened 2022-01-30
functions: handle gzip compressed kernels in kver_generic

jannau:
Instead of this hard-coded gzip fallback solution I could also add compression detection via `file` or `od`.


jannau:
Closing for the functionally equivalent #112

PR #89, opened 2022-03-11
add MODULES_DECOMPRESS switch

tpowa:
Third one:
add MODULES_DECOMPRESS switch feature
https://bugs.archlinux.org/task/72882

### DIFF
--- a/functions
+++ b/functions
@@ -860,7 +860,7 @@ install_modules() {
     if [[ "$MODULES_DECOMPRESS" = yes ]]; then
         # unzip modules prior to recompression
         for m in "$@"; do
-            case $m in
+            case "$m" in
                 *.xz)
                     xz_comp+=("$moduledest/kernel/${m##*/}")
                     ;;
--- a/functions
+++ b/functions
@@ -846,7 +846,7 @@ try_enable_color() {
 
 install_modules() {
     local m moduledest=$BUILDROOT/lib/modules/$KERNELVERSION
-    local -a xz_comp gz_comp zst_comp
+    MODULES_DECOMPRESS="yes" && local -a xz_comp gz_comp zst_comp
 
     [[ $KERNELVERSION == none ]] && return 0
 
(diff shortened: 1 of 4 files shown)


mkinitcpio.conf
```diff
@@ -65,3 +65,9 @@ HOOKS=(base udev autodetect modconf block filesystems keyboard fsck)
 # COMPRESSION_OPTIONS
 # Additional options for the compressor
 #COMPRESSION_OPTIONS=()
+
+# MODULES_DECOMPRESS
+# Decompress kernel modules during initramfs creation.
+# Enable to speedup boot process, disable to save RAM
+# during early userspace. Switch (yes/no) default=yes
+#MODULES_DECOMPRESS=""
```
nl6720, line 73: How about showing the default value inline instead of in the accompanying comment:

```suggestion
# during early userspace. Switch (yes/no).
#MODULES_DECOMPRESS="yes"
```

functions
```diff
@@ -846,7 +846,7 @@ try_enable_color() {
 
 install_modules() {
     local m moduledest=$BUILDROOT/lib/modules/$KERNELVERSION
-    local -a xz_comp gz_comp zst_comp
+    MODULES_DECOMPRESS="yes" && local -a xz_comp gz_comp zst_comp
```
nl6720, line 849: Combining these lines doesn't feel right to me.
```suggestion
    local -a xz_comp gz_comp zst_comp
    MODULES_DECOMPRESS="yes"
```

Foxboron: Is it intentional that `install_modules` is forced to decompress? I don't quite see why we have the config option if we ignore it.

mkinitcpio
```diff
@@ -639,6 +639,8 @@ if [[ $KERNELVERSION != 'none' ]]; then
     [[ -d $_d_kmoduledir ]] || die "'$_d_kmoduledir' is not a valid kernel module directory"
 fi
 
+[[ -z $MODULES_DECOMPRESS ]] && MODULES_DECOMPRESS=yes
```
Foxboron, line 642:
```suggestion
MODULES_DECOMPRESS=${MODULES_DECOMPRESS:-"yes"}
```

functions
tpowa: Foxboron you are right, nl6720 proposal is not a fix.

Foxboron: I reckon just removing the line is the correct approach here :)

functions
```diff
@@ -857,20 +857,22 @@ install_modules() {
...
-                ;;
-            *.gz)
-                gz_comp+=("$moduledest/kernel/${m##*/}")
-                ;;
-            *.zst)
-                zst_comp+=("$moduledest/kernel/${m##*/}")
-                ;;
-        esac
-    done
+    if [[ $MODULES_DECOMPRESS == yes ]]; then
```
IsaacG, line 860:
```suggestion
    if [[ "$MODULES_DECOMPRESS" = yes ]]; then
```

functions
```diff
@@ -857,20 +857,22 @@ install_modules() {
...
+        # unzip modules prior to recompression
+        for m in "$@"; do
+            case $m in
```
IsaacG, line 863:
```suggestion
            case "$m" in
```

functions
nl6720: No! `=` is the POSIX syntax. Bashisms, such as `==`, should be prefered.

IsaacG: 1. Is there some sort of style guide which recommends using always using bashisms? I've never heard any style guide dictate any such rationale.
2. We can probably all agree you shouldn't use `[[ ! a == b ]]` over [[ a != b ]]` just to use more bashisms, right? bashisms are good to use, when there is value in their use; they shouldn't be used just to use them. That's cargo culting.
3. `[[ a = b ]]` _is_ a bashism. `=` without context is just a char. `[[` is a bashism, so all `[[` operators are bashisms.
4. `help [[` says `When the `==' and `!=' operators are used, the string to the right of the operator is used as a pattern and pattern matching is performed.` Compare to `help test` and `STRING1 = STRING2  True if the strings are equal.` which is different.

nl6720: The project guidelines says to generally prefer bash syntax. And I agree.

https://github.com/archlinux/mkinitcpio/blob/7bfe4861eacb3bf6cb70d9a17a0262542733a8ed/HACKING#L7-L11

Since `[[` is different from `[`/`test`, `help test` doesn't apply to it. `[[` will always perform pattern matching.

functions
nl6720:
```suggestion
    local -a xz_comp gz_comp zst_comp
```

nl6720:
Merged in https://github.com/archlinux/mkinitcpio/commit/87ae9cbab2f153406483e54ec9b4594c3ebf0988.

PR #90, opened 2022-03-11
fix reproducibility, in case of /var/cache directory exists and is included in initramfs

tpowa:
4th:
Remove not needed and not reproducable cache file
https://bugs.archlinux.org/task/73817

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -709,6 +709,8 @@ trap -- ERR
 
 # this is simply a nice-to-have -- it doesn't matter if it fails.
 ldconfig -r "$BUILDROOT" &>/dev/null
+# remove /var/cache/ldconfig/aux-cache for reproducability
+rm -f -- "$BUILDROOT/var/cache/ldconfig/aux-cache"
 
 # Set umask to create initramfs images and EFI images as 600
 umask 077


mkinitcpio
```diff
@@ -709,6 +709,8 @@ trap -- ERR
 
 # this is simply a nice-to-have -- it doesn't matter if it fails.
 ldconfig -r "$BUILDROOT" &>/dev/null
+# remove /var/cache/ldconfig/aux-cache for reproducability
```
orhun, line 712: Hey! There is a typo in this comment. ("reproducability" -> "reproducibility")

mkinitcpio
```diff
@@ -709,6 +709,8 @@ trap -- ERR
...
+[[ -e $BUILDROOT/var/cache/ldconfig/aux-cache ]] && rm $BUILDROOT/var/cache/ldconfig/aux-cache
```
nl6720, line 713: You can just avoid the test and use `rm -f`:
```suggestion
rm -f $BUILDROOT/var/cache/ldconfig/aux-cache
```

nl6720 (changes requested):

mkinitcpio
tpowa: Thanks fixed.

Foxboron:
I'm still not really sold on this change. `/var/cache` doesn't belong in an initramfs and I struggle to see a good reason for us to be patching this.

tpowa:
You are free to include anything,,, it's a real bug. pacman's default cache sits there.

tpowa:
To summarize again the issue:
- ldconfig which is called at the end of mkinitcpio, will create a cache file in /var/cache/ldconfig/aux-cache, if /var/cache directory exists.
- If for whatever reason /var/cache (the directory only!) was added to initramfs (in my case to allow full pacman support in initramfs of archboot), you will never be able to get a reproducable initramfs.
- By rejecting this request you break reproducable builds for users

- Up to the point, there was no argument given yet why this file should be in initramfs /var/cache/ldconfig/aux-cache?
- By rejecting this, you imply that this file is needed in initramfs.
- ldconfig is using an aux-cache to speed up the ld.so.cache update, which will not happen in normal initramfs usage.

Please give me valid arguments why this oneliner shouldn't be part of the software, I wrote 17 years ago in mind of giving users full power over their early userspace.

mkinitcpio
```diff
@@ -709,6 +709,8 @@ trap -- ERR
 
 # this is simply a nice-to-have -- it doesn't matter if it fails.
 ldconfig -r "$BUILDROOT" &>/dev/null
+# remove /var/cache/ldconfig/aux-cache for reproducibility
+rm -f "$BUILDROOT"/var/cache/ldconfig/aux-cache
```
nl6720, line 713: Terminate option processing:

```suggestion
rm -f -- "$BUILDROOT"/var/cache/ldconfig/aux-cache
```

nl6720 (changes requested):

mkinitcpio
nl6720, line 713:
```suggestion
rm -f -- "$BUILDROOT/var/cache/ldconfig/aux-cache"
```

nl6720 (changes requested):

PR #93, opened 2022-03-14
Improve bash syntax in mkinitcpio, lsinitcpio and functions

nl6720:
Quote, quote and quote some more.

Most of the issues were found with shellcheck.

### DIFF
--- a/functions
+++ b/functions
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 parseopts() {
-    local opt= optarg= i= shortopts=$1
+    local opt='' optarg='' i='' shortopts="$1"
     local -a longopts=() unused_argv=()
 
     shift
@@ -14,18 +14,18 @@ parseopts() {
     longoptmatch() {
         local o longmatch=()
         for o in "${longopts[@]}"; do
-            if [[ ${o%:} = "$1" ]]; then
+            if [[ "${o%:}" == "$1" ]]; then
                 longmatch=("$o")
                 break
             fi
-            [[ ${o%:} = "$1"* ]] && longmatch+=("$o")
+            [[ "${o%:}" == "$1"* ]] && longmatch+=("$o")
         done
 
-        case ${#longmatch[*]} in
+        case "${#longmatch[*]}" in
             1)
                 # success, override with opt and return arg req (0 == none, 1 == required)
-                opt=${longmatch%:}
-                if [[ $longmatch = *: ]]; then
+                opt="${longmatch%:}"
+                if [[ "${longmatch[*]}" == *: ]]; then
                     return 1
                 else
                     return 0
@@ -42,7 +42,7 @@ parseopts() {
     }
 
     while (( $# )); do
-        case $1 in
+        case "$1" in
             --) # explicit end of options
                 shift
                 break
@@ -60,7 +60,7 @@ parseopts() {
 
                     OPTRET+=("-$opt")
                     # option requires optarg
-                    if [[ $shortopts = *$opt:* ]]; then
+                    if [[ "$shortopts" == *"${opt}:"* ]]; then
                         # if we're not at the end of the option chunk, the rest is the optarg
                         if (( i < ${#1} - 1 )); then
                             OPTRET+=("${1:i+1}")
@@ -138,8 +138,9 @@ kver_x86() {
     # to the version string by reading 2 bytes out of image at at
     # address 0x20E. this leads us to a string of, at most, 128 bytes.
     # read the first word from this string as the kernel version.
-    local kver offset=$(hexdump -s 526 -n 2 -e '"%0d"' "$1")
-    [[ $offset = +([0-9]) ]] || return 1
+    local kver offset
+    offset="$(hexdump -s 526 -n 2 -e '"%0d"' "$1")"
+    [[ "$offset" == +([0-9]) ]] || return 1
 
     read kver _ < \
         <(dd if="$1" bs=1 count=127 skip=$(( offset + 0x200 )) 2>/dev/null)
@@ -151,7 +152,7 @@ detect_compression() {
     local bytes
 
     read -rd '' bytes < <(hexdump -n 6 -e '"%c"' "$1")
-    case $bytes in
+    case "$bytes" in
         $'\xfd7zXZ')
             echo 'xz'
             return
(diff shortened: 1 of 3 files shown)


functions
```diff
@@ -14,18 +14,18 @@ parseopts() {
     longoptmatch() {
         local o longmatch=()
         for o in "${longopts[@]}"; do
-            if [[ ${o%:} = "$1" ]]; then
+            if [[ "${o%:}" == "$1" ]]; then
```
IsaacG, line 17: The `==` allows pattern matching (according to `man bash`). It's not needed for string equality.

functions
```diff
@@ -60,7 +60,7 @@ parseopts() {
 
                     OPTRET+=("-$opt")
                     # option requires optarg
-                    if [[ $shortopts = *$opt:* ]]; then
+                    if [[ "$shortopts" == *$opt:* ]]; then
```
IsaacG, line 63:
```suggestion
                    if [[ "$shortopts" = *"$opt:"* ]]; then
```

functions
```diff
@@ -254,7 +256,7 @@ index_of() {
     local item=$1; shift
 
     for (( _idx=1; _idx <= $#; _idx++ )); do
-        if [[ $item = ${!_idx} ]]; then
+        if [[ $item == ${!_idx} ]]; then
```
IsaacG, line 259: The `==` isn't needed ... but the quotes were not added

functions
```diff
@@ -310,10 +312,10 @@ all_modules() {
     #   -f FILTER: ERE to filter found modules
 
     local -i count=0
-    local mod= OPTIND= OPTARG= filter=()
+    local mod='' OPTIND='' OPTARG='' filter=()
 
     while getopts ':f:' flag; do
-        case $flag in f) filter+=("$OPTARG") ;; esac
+        case "$flag" in f) filter+=("$OPTARG") ;; esac
```
IsaacG, line 318:
```suggestion
        [[ "$flag" = "f" ]] && filter+=("$OPTARG")
```

functions
```diff
@@ -388,17 +390,17 @@ add_module() {
...
     #   $1: module name
 
-    local target= module= softdeps= deps= field= value= firmware=()
+    local target='' module='' softdeps='' deps='' field='' value='' firmware=()
     local ign_errors=0 found=0
 
     [[ $KERNELVERSION == none ]] && return 0
 
-    if [[ $1 = *\? ]]; then
+    if [[ $1 == *\? ]]; then
```
IsaacG, line 398:
```suggestion
    if [[ "$1" == *\? ]]; then
```

functions
```diff
@@ -410,7 +412,7 @@ add_module() {
                 # it might be reported as "(builtin)"). We'll defer actually
                 # checking whether or not the file exists -- any errors can be
                 # handled during module install time.
-                if [[ $value = /* ]]; then
+                if [[ $value == /* ]]; then
```
IsaacG, line 415:
```suggestion
                if [[ "$value" == /* ]]; then
```

functions
```diff
@@ -500,19 +502,19 @@ add_dir() {
...
 
-    if [[ -d $BUILDROOT$1 ]]; then
+    if [[ -d "${BUILDROOT}${1}" ]]; then
         # ignore dir already exists
         return 0
     fi
 
     quiet "adding dir: %s" "$path"
-    command install -dm$mode "$BUILDROOT$path"
+    command install -dm"${mode}" "${BUILDROOT}${path}"
```
IsaacG, line 517: Inconsistent use of braces. Either add it to all the variables or don't add it anywhere. Adding it in a few places makes the style inconsistent.

functions
```diff
@@ -808,7 +811,7 @@ run_build_hook() {
     fi
 
     if ! declare -f build >/dev/null; then
-        error 'Hook '$script' has no build function'
+        error 'Hook '"$script"' has no build function'
```
IsaacG, line 814:
```suggestion
        error "Hook $script"' has no build function"
```

functions
```diff
@@ -829,26 +832,27 @@ run_build_hook() {
...
+        _color_red="$_color_bold$(tput setaf 1)"
+        _color_yellow="$_color_bold$(tput setaf 3)"
     fi
 }
 
 install_modules() {
-    local m moduledest=$BUILDROOT/lib/modules/$KERNELVERSION
+    local m moduledest
     local -a xz_comp gz_comp zst_comp
+    moduledest="$BUILDROOT/lib/modules/$KERNELVERSION"
```
IsaacG, line 853: This can be in the `local` line as it does not execute anything.

mkinitcpio
```diff
@@ -526,7 +527,7 @@ while :; do
         -U|--uefi)
             shift
             [[ -d $1 ]] && die "Invalid image path -- must not be a directory"
```
IsaacG, line 529:
```suggestion
            [[ -d "$1" ]] && die "Invalid image path -- must not be a directory"
```

mkinitcpio
```diff
@@ -551,43 +552,43 @@ while :; do
         --splash)
             shift
             [[ -f $1 ]] || die "Invalid file -- must be a file"
```
IsaacG, line 554:
```suggestion
            [[ -f "$1" ]] || die "Invalid file -- must be a file"
```

functions
nl6720: Both `==` and `=` compare strings. The only difference is that `=` is the POSIX compatible syntax, while `==` is a bashism. Since it's a bash scripts, we should prefer bashisms.

functions
nl6720: The `error` function uses `prinft`, so this will work:

```suggestion
        error 'Hook %s has no build function' "$script"
```

functions
nl6720: I guess I was too afraid of breaking things.

functions
nl6720: The way it's written makes me think that it's intentional, but besides future proofing in case another parameter is added, I can't imaging why. :thinking:

functions
IsaacG: Most style guides push for consistent use, one way or another, since consistent code is good code.

In this particular line, they aren't needed anywhere, which is how the code was working prior to this change.

nl6720: It just feel dirty to me to place variables inline without enclosing them.

klausenbusk:
Migrated to to: https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/merge_requests/93.

PR #96, opened 2022-03-17
Preserve kernel structure, remove awk hack on install_modules

tpowa:
- messing around with modules is not really needed 
- modules.order does not include aliases or double entries
- if a module alias should conflict, it's already a corner case which should be solved in MODULES='' array or by the other kernel methods documented to stop loading a certain conflicting module. (module blacklisting).

### DIFF
--- a/functions
+++ b/functions
@@ -855,19 +855,18 @@ install_modules() {
         return 0
     fi
 
-    cp "$@" "$moduledest/kernel"
-
-    # unzip modules prior to recompression
     for m in "$@"; do
+        install -D -m644 "$m" "$BUILDROOT/$m"
+        # unzip modules prior to recompression
         case $m in
             *.xz)
-                xz_comp+=("$moduledest/kernel/${m##*/}")
+                xz_comp+=("$BUILDROOT/$m")
                 ;;
             *.gz)
-                gz_comp+=("$moduledest/kernel/${m##*/}")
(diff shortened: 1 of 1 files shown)


nl6720 (changes requested):
Please rebase to get rid of the `Merge branch 'master' into feature-6` commit.

functions
```diff
@@ -877,14 +876,6 @@ install_modules() {
 
     msg "Generating module dependencies"
     install -m644 -t "$moduledest" "$_d_kmoduledir"/modules.{builtin,builtin.modinfo,order}
```
nl6720, line 878: Not directly related to this PR, but how do we know which files we need?
The modules directory has: `modules.alias`, `modules.alias.bin`, `modules.builtin`, `modules.builtin.alias.bin`, `modules.builtin.bin`, `modules.builtin.modinfo`, `modules.dep`, `modules.dep.bin`, `modules.devname`, `modules.order`, `modules.symbols`, `modules.symbols.bin`, `modules.softdep`.

nl6720:
Commit https://github.com/archlinux/mkinitcpio/commit/2e01e8a3a7e81eeb52799a8dcd84edd4b1c7d78f merged with https://github.com/archlinux/mkinitcpio/commit/67fe3bf7087b1e0a38a4bee8a0c49642a59f1169.

PR #78, opened 2021-12-15
presets: make {default,fallback}_image optional when using *_efi_image

gdamjan:
fixes #77

PR #99, opened 2022-04-07
preset: Change unified image generation options

### LINKED ISSUE #77: make default_image in presets optional when UEFI is used
When building unified UEFI images with `mkinitcpio --uefi /tmp/test.efi`, mkinitcpio will create (and later remove) a **temporary** initcpio image (`/tmp/mkinitcpio.dJk7yP`).

On the other hand, in preset files, the `default_image` parameter is mandatory, even though `default_efi_image="/boot/EFI/Linux/archlinux-linux.efi"` is used.

For ex:
```
==> Building image from preset: /etc/mkinitcpio.d/linux.preset: 'default'
==> WARNING: No image file specified. Skipping image `default'
```

I suggest that if `default_efi_image=` is set, `default_image=` should be optional. Thoughts?

Molytho:
- Setting ALL_efi_image is nonsense (like it is for ALL_image) so don't allow it
- Allow omitting *_image when *_efi_image is set

fixes #77
supersedes #78

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -399,22 +399,22 @@ process_preset() (
             continue
         fi
 
+        preset_efi_image=${p}_efi_image
+        if [[ ${!preset_efi_image} ]]; then
+            preset_cmd+=(-U "${!preset_efi_image}")
+        fi
+
         preset_image=${p}_image
         if [[ ${!preset_image} ]]; then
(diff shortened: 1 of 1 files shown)


Foxboron:
Merged as https://github.com/archlinux/mkinitcpio/commit/4f9896c3549c151cfd495ba24268bd977f7993e5

PR #100, opened 2022-04-07
Fix issues related to uefistub argument

Molytho:
- The uefistub argument is broken because of typos and copy-paste issues


Foxboron:
This is fixed by https://github.com/archlinux/mkinitcpio/pull/80 as well

Molytho:
Sorry, missed this. As it does exactly the same I will close it

PR #97, opened 2022-03-19
Variable typo correction

Bischnu:
I corrected a typo in a $uefistub variable used in the build_uefi function.

Edit: corrected another typo.

PR #101, opened 2022-04-15
fixed empty uefistub detection

RichardKraus:
In the current state if the uefistub arg is empty and the loop fails to find a stub it will still continue execution as the check for an empty uefistub is done in the elif.

This leads to nondescript error messages like:
```
objcopy: '': No such file
objcopy: --change-section-vma .splash=0x0000000000040000 never used
objcopy: --change-section-vma .initrd=0x0000000003000000 never used
objcopy: --change-section-vma .linux=0x0000000002000000 never used
objcopy: --change-section-vma .cmdline=0x0000000000030000 never used
objcopy: --change-section-vma .osrel=0x0000000000020000 never used
==> ERROR: UEFI executable generation FAILED
```
on systems where systemd-stub is not installed

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -278,7 +278,7 @@ build_uefi(){
             fi
         done
     fi
-    if [[ ! -f "$uefisub" ]]; then
+    if [[ ! -f "$uefistub" ]]; then
         error "UEFI stub '%s' not found" "$uefistub"
         return 1
     fi
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -277,7 +277,8 @@ build_uefi(){
                 break
             fi
         done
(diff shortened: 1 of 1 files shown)


Bischnu:
Hello, I do not know well how conflicting branches are resolved, there is a typo in uefistub variable name on the line 280/281 which was corrected in pull requests #80 and #97. Maybe it should be edited here as well?

RichardKraus:
oh wow i didnt even see that and just edited the line ':D

Foxboron:
Squashed and merged as https://github.com/archlinux/mkinitcpio/commit/1d14c7329ff414230dd8b1945d5d57097dd07512

Issue #102, opened 2022-04-22
Syntax error in /usr/lib/initcpio/functions


Strahinja:
Currently, the file `functions` in the root of the repository gives an error when that file is sourced:
```
bash-5.1.16$ source /usr/lib/initcpio/functions
bash: /usr/lib/initcpio/functions: line 682: syntax error near unexpected token `('
bash: /usr/lib/initcpio/functions: line 682: `                RUN@({program}|+)|IMPORT{program}|ENV{REMOVE_CMD})'
```
It seems that a case pattern contains parentheses, which is unsupported in shell pattern matching and produces an error. I suggest expanding the [condition][1] as follows:
```
RUN{program}|RUN+|IMPORT{program}|ENV{REMOVE_CMD})
```

**Edit:** If the intent was to use `extglob`-dependent `@()` construct, then `extglob` should be explicitly set, at least during the execution of `add_udev_rule`. However, that wouldn't respect user's choice for the option.

[1]: https://github.com/archlinux/mkinitcpio/blob/master/functions#L682

PR #103, opened 2022-04-22
Expand @() as it fails when extglob is not set

Strahinja:
Currently, as per #102,  sourcing /usr/lib/initcpio/functions when `extglob` is not set in Bash produces an error. This PR expands the case pattern so `@()` extended pattern is not used.

The alternative would be to force `shopt -s extglob` at the top of the file, but that wouldn't respect user's choice for this option.

### DIFF
--- a/functions
+++ b/functions
@@ -679,7 +679,7 @@ add_udev_rule() {
         for pair in "${rule[@]}"; do
             IFS=' =' read -r key value <<< "$pair"
             case $key in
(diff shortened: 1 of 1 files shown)


Foxboron:
Merged as https://github.com/archlinux/mkinitcpio/commit/3615e019b8a09a218bf02c4d8f6fb655c7805ac2

Thanks!

PR #109, opened 2022-07-03
Fix the warning about missing modules.builtin.modinfo

yan12125:
> depmod: WARNING: could not open modules.builtin.modinfo at /tmp/mkinitcpio.EftMSK/root/lib/modules/5.15.52-1-lts: No such file or directory

kmod wants modules.builtin.modinfo since v30 [1]

[1] https://git.kernel.org/pub/scm/utils/kernel/kmod/kmod.git/commit/?id=0246e06340df292b5dda4bc00e24cc9ae894e881

### DIFF
--- a/functions
+++ b/functions
@@ -876,7 +876,7 @@ install_modules() {
     (( ${#zst_comp[*]} )) && zstd -d --rm -q "${zst_comp[@]}"
 
     msg "Generating module dependencies"
-    install -m644 -t "$moduledest" "$_d_kmoduledir"/modules.builtin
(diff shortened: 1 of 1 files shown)


heftig:
While you're at it, you should also add `modules.order` so we have the complete set, though it's not strictly required.

yan12125:
Thanks! Updated.

yan12125:
Probably related to this - modprobe uses modules.dep.bin generated by depmod.
```
modules.order
--------------------------------------------------
This file records the order in which modules appear in Makefiles. This
is used by modprobe to deterministically resolve aliases that match
multiple modules.
```
(https://www.kernel.org/doc/Documentation/kbuild/kbuild.txt)

PR #112, opened 2022-08-22
Support gzip compressed linux kernels for RISC-V kernels

felixonmars:
Make the fallback kver scraper test for gzip and use `zcat` for gzip compressed kernels.

This has been tested on qemu-system-riscv64 and the generated initramfs works pretty well.

### DIFF
--- a/functions
+++ b/functions
@@ -147,6 +147,65 @@ kver_x86() {
     printf '%s' "$kver"
 }
 
+detect_compression() {
+    local bytes
+
+    read -rd '' bytes < <(hexdump -n 6 -e '"%c"' "$1")
+    case $bytes in
+        $'\xfd7zXZ')
+            echo 'xz'
+            return
+            ;;
+    esac
+
+    read -rd '' bytes < <(hexdump -n 4 -e '"%c"' "$1")
+    if [[ $bytes = $'\x89LZO' ]]; then
+        echo 'lzop'
+        return
+    fi
+
+    read -rd '' bytes < <(hexdump -n 2 -e '"%x"' "$1")
+    if [[ $bytes = '8b1f' ]]; then
+        echo 'gzip'
+        return
+    fi
+
+    read -rd '' bytes < <(hexdump -n 4 -e '"%x"' "$1")
+    case $bytes in
+        184d2204)
+            error 'Newer lz4 stream format detected! This may not boot!'
+            echo 'lz4'
+            return
+            ;;
+        184c2102)
+            echo 'lz4 -l'
+            return
+            ;;
+        fd2fb528)
+            echo 'zstd'
+            return
+            ;;
+    esac
+
+    read -rd '' bytes < <(hexdump -n 3 -e '"%c"' "$1")
+    if [[ $bytes == 'BZh' ]]; then
+        echo 'bzip2'
+        return
+    fi
+
+    # lzma detection sucks and there's really no good way to
+    # do it without reading large portions of the stream. this
+    # check is good enough for GNU tar, apparently, so it's good
+    # enough for me.
+    read -rd '' bytes < <(hexdump -n 3 -e '"%x"' "$1")
+    if [[ $bytes = '5d' ]]; then
+        echo 'lzma'
+        return
+    fi
+
+    # out of ideas, assuming uncompressed
+}
+
 kver_generic() {
     # For unknown architectures, we can try to grep the uncompressed or gzipped
     # image for the boot banner.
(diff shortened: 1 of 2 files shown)


functions
```diff
@@ -148,17 +148,19 @@ kver_x86() {
...
+    # This should work at least for ARM when run on /boot/Image, or RISC-V on
+    # gzipped /boot/vmlinuz-linuz. On other architectures it may be worth trying
+    # rather than bailing, and inform the user if none was found.
 
     # Loosely grep for `linux_banner`:
     # https://elixir.bootlin.com/linux/v5.7.2/source/init/version.c#L46
-    local kver=
+    local kver= reader=cat
+    
+    [[ $(file -b --mime-type "$1") == 'application/gzip' ]] && reader=zcat
```
Foxboron, line 161: I'm unsure about the `file` usage. It seems like we use hexdump in `detect_filetype`. We should probably try using that instead of reimplementing this with a new dependency?

felixonmars: OK. I have moved the compression detection part out of `detect_filetype` into a new function `detect_compression`, and shared its usage between `lsinitcpio` and `mkinitcpio`.

Foxboron: Cool. I would maybe use a case statement for selecting the correct reader thing. But this is better i think :)

functions
```diff
@@ -147,18 +147,79 @@ kver_x86() {
...
+detect_compression() {
+    local bytes
+
+    read -rd '' bytes < <(hexdump -n 6 -e '"%c"' "$1")
+    case $bytes in
+        $'\xfd7zXZ')
+            echo 'xz'
+            return
+            ;;
+    esac
```
IsaacG, line 159:
```suggestion
    if [[ $bytes = $'\xfd7zXZ' ]]; then
            echo 'xz'
            return
    fi
```

functions
```diff
@@ -147,18 +147,79 @@ kver_x86() {
...
+    fi
+
+    read -rd '' bytes < <(hexdump -n 2 -e '"%x"' "$1")
+    if [[ $bytes = '8b1f' ]]; then
+        echo 'gzip'
+        return
+    fi
+
+    read -rd '' bytes < <(hexdump -n 4 -e '"%x"' "$1")
+    case $bytes in
```
IsaacG, line 174:
```suggestion
    case "$bytes" in
```

functions
```diff
@@ -147,18 +147,79 @@ kver_x86() {
...
+        return
+    fi
+
+    read -rd '' bytes < <(hexdump -n 2 -e '"%x"' "$1")
+    if [[ $bytes = '8b1f' ]]; then
+        echo 'gzip'
+        return
+    fi
+
+    read -rd '' bytes < <(hexdump -n 4 -e '"%x"' "$1")
```
IsaacG, line 173: Can the `lzop` and these be combined of must they use different dump formats?

functions
```diff
@@ -147,18 +147,79 @@ kver_x86() {
...
+    # This should work at least for ARM when run on /boot/Image, or RISC-V on
+    # gzipped /boot/vmlinuz-linuz. On other architectures it may be worth trying
+    # rather than bailing, and inform the user if none was found.
 
     # Loosely grep for `linux_banner`:
     # https://elixir.bootlin.com/linux/v5.7.2/source/init/version.c#L46
-    local kver=
+    local kver= reader=cat
+    
+    [[ $(detect_compression "$1") == 'gzip' ]] && reader=zcat
```
IsaacG, line 220:
```suggestion
    [[ $(detect_compression "$1") = 'gzip' ]] && reader=zcat
```

functions
```diff
@@ -147,18 +147,79 @@ kver_x86() {
...
 
-    read _ _ kver _ < <(grep -m1 -aoE 'Linux version .(\.[-[:alnum:]]+)+' "$1")
+    read _ _ kver _ < <($reader "$1" | grep -m1 -aoE 'Linux version .(\.[-[:alnum:]]+)+')
```
IsaacG, line 222:
```suggestion
    read _ _ kver _ < <("$reader" "$1" | grep -m1 -aoE 'Linux version .(\.[-[:alnum:]]+)+')
```

functions
nl6720: No. We don't need to adhere to POSIX sh syntax in a bash script.

felixonmars:
While I really appreciate the review work, I don't think combining style changes in a functional change PR is a good idea. I would actually recommend style changes to be in a separated PR and update at least a whole file at a time.

functions
felixonmars: I don't really know any of these and the code was moved from `lsinitcpio` to here for function sharing. Probably optimization of this function should be in a separate PR.

Foxboron:
I agree, we are moving stuff around a bit so we can use the function in both commands. I think this iteration is fine :)

felixonmars:
Conflict after #75 is solved.

nl6720 (changes requested):
Please rebase to get rid of the `Merge branch 'master' into patch-1` commit.

nl6720 (approved):

PR #113, opened 2022-08-22
install/block: include spi modules

felixonmars:
SPI is used to provide access to mmc devices on some development boards (for example, HiFive Unmatched). The board doesn't boot from SD card properly without them included.

The whole `drivers/spi` folder is 380KiB in size, uncompressed, it shouldn't be a lot to add :)

### DIFF
--- a/install/block
+++ b/install/block
@@ -20,6 +20,9 @@ build() {
 
     # mmc
     add_checked_modules '/(drivers/mmc|tifm_)'
+    
+    # spi
+    add_checked_modules '/drivers/spi/'
 
     # virtio
     add_checked_modules 'virtio'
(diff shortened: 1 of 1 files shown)


nl6720:
Is this is about [MMC in SPI mode](http://elm-chan.org/docs/mmc/mmc_e.html#spimode)? If so, IMHO, the comment should mention it.

nl6720 (changes requested):
Please squash.

Foxboron:
Merged and squashed here https://github.com/archlinux/mkinitcpio/commit/ae34922e47fb0a049a9c1c7a5d9c484b210b6c85

Issue #64, opened 2021-08-18
Overwrite-by-rename instead of truncating images for more crash resilience


Artefact2:
The current code truncates existing images when generating the initramfs. This can/will lead to an unbootable system if a crash happens shortly after having run mkinitcpio.

I propose using an atomic overwrite-by-rename instead, so that the initramfs file is always consistent. Simple proof of concept follows below.

~~~
diff --git a/mkinitcpio.old b/mkinitcpio.new
index 46360ef..c9801dc 100644
--- a/mkinitcpio.old
+++ b/mkinitcpio.new
@@ -226,7 +226,9 @@ build_image() {
             sort -z |
             LANG=C bsdtar --null -cnf - -T - |
             LANG=C bsdtar --uid 0 --gid 0 --null -cf - --format=newc @- |
-            $compress "${COMPRESSION_OPTIONS[@]}" > "$out"
+            $compress "${COMPRESSION_OPTIONS[@]}" > "$out".tmp
+    sync -d "$out.tmp"
+    mv -f "$out".tmp "$out"
 
     pipestatus=("${PIPESTATUS[@]}")
     pipeprogs=('find' 'sort' 'bsdtar (step 1)' 'bsdtar (step 2)' "$compress")

~~~

christian-heusel:
Should the generated temporary files go into the output dir as proposed above or should we use a tempdir as created by `mktemp`?

nl6720:
Using a tempdir wouldn't solve the issue. It needs to be on the same file system as the file it replaces.

PR #124, opened 2022-10-22
ensure consistency for the generated image

christian-heusel:
This just applies the proposed fix from @Artefact2 in the linked issue.
Should we make this behaviour more configurable as it increases the space requirements, which could be a breaking change for people with full `/boot`?

fixes https://github.com/archlinux/mkinitcpio/issues/64

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -238,7 +238,9 @@ build_image() {
             sort -z |
             LANG=C bsdtar --uid 0 --gid 0 --null -cnf - -T - |
             LANG=C bsdtar --null -cf - --format=newc @- |
-            $compress "${COMPRESSION_OPTIONS[@]}" > "$out"
+            $compress "${COMPRESSION_OPTIONS[@]}" > "$out".tmp
+    sync -d "$out".tmp
+    mv -f "$out".tmp "$out"
 
     pipestatus=("${PIPESTATUS[@]}")
     pipeprogs=('find' 'sort' 'bsdtar (step 1)' 'bsdtar (step 2)' "$compress")
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -231,13 +231,12 @@ build_image() {
     if [[ -f "$out" ]]; then
         local curr_size space_left_on_device
 
-        curr_size=$(stat --format="%s" $out)
-        space_left_on_device=$(($(stat -f --format="%a*%S" $out)))
+        curr_size="$(stat --format="%s" "$out")"
+        space_left_on_device="$(($(stat -f --format="%a*%S" "$out")))"
 
         # check if there is enough space on the device to write the image to a tempfile, fallback otherwise
-        if (( $(($curr_size*2)) < $space_left_on_device )); then
-            compressout="$out".tmp
-        fi
+        # this assumes that the new image is not more than 1¼ times the size of the old one
+        (( $(($curr_size + ($curr_size/4))) < $space_left_on_device )) && compressout="$out".tmp
     fi
 
     pushd "$BUILDROOT" >/dev/null
@@ -251,12 +250,6 @@ build_image() {
             LANG=C bsdtar --uid 0 --gid 0 --null -cnf - -T - |
             LANG=C bsdtar --null -cf - --format=newc @- |
             $compress "${COMPRESSION_OPTIONS[@]}" > "$compressout"
-
-    # sync and rename as we wrote to a tempfile
-    if [[ "$compressout" != "$out" ]]; then
-        sync -d "$compressout"
-        mv -f "$compressout" "$out"
-    fi
 
     pipestatus=("${PIPESTATUS[@]}")
     pipeprogs=('find' 'sort' 'bsdtar (step 1)' 'bsdtar (step 2)' "$compress")
@@ -279,6 +272,12 @@ build_image() {
         return 1
     elif (( _builderrors == 0 )); then
         msg "Image generation successful"
+    fi
+
+    # sync and rename as we only wrote to a tempfile so far to ensure consistency
+    if [[ "$compressout" != "$out" ]]; then
+        sync -d -- "$compressout"
+        mv -f -- "$compressout" "$out"
     fi
 }
 
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -207,7 +207,7 @@ compute_hookset() {
 }
 
 build_image() {
-    local out=$1 compress=$2 errmsg pipestatus
+    local out=$1 compressout=$1 compress=$2 errmsg pipestatus
 
     case $compress in
         cat)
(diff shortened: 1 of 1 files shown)


christian-heusel:
I could also add Co-Authorship for the person that opened the issue if wanted!

Artefact2:
It's a trivial fix, I don't mind. Happy to see it fixed in mainline :)

Foxboron (approved):

nl6720:
> it increases the space requirements

IMHO this is quite a big issue. People with 100 MiB `/boot` already exceed space requirements frequently enough.

The [suggested `/boot` size in the wiki](https://wiki.archlinux.org/title/Partitioning#/boot) would need updating too.

christian-heusel:
> it increases the space requirements

It increases the space requirements by `max(img_filesizes)` as the rename is happening after each image, so we get to a new needed minimum amount of `sum(img_filesizes) + max(img_filesizes)`.
For the two machines i am testing this with this means the increase is +34M because of the size of `initramfs-linux-fallback.img`.

How big can these get?

nl6720:
Generally, the more crap you add, the bigger it gets.
I myself have a 49 MiB `initramfs-linux-fallback.img`, but I'm using 
```bash
COMPRESSION="xz"
COMPRESSION_OPTIONS=(-9e)
```
With the default compressions options, it's 66 MiB for me.

Artefact2:
The initramfses can get quite large, especially in NVIDIA users that want early kms (80-ish megs last time I checked).

nl6720 (changes requested):
I think there should be a fallback to truncating in case there's not enough free space.

mkinitcpio
```diff
@@ -228,6 +228,18 @@ build_image() {
...
+    if [[ -f "$out" ]]; then
+        local curr_size space_left_on_device
+
+        curr_size=$(stat --format="%s" $out)
+        space_left_on_device=$(($(stat -f --format="%a*%S" $out)))
+
+        # check if there is enough space on the device to write the image to a tempfile, fallback otherwise
+        if (( $(($curr_size*2)) < $space_left_on_device )); then
+            compressout="$out".tmp
+        fi
```
nl6720, line 240:
```suggestion
        (( $((curr_size*2)) < space_left_on_device )) && compressout="$out".tmp
```

mkinitcpio
```diff
@@ -228,6 +228,18 @@ build_image() {
             ;;
     esac
 
+    if [[ -f "$out" ]]; then
+        local curr_size space_left_on_device
+
+        curr_size=$(stat --format="%s" $out)
+        space_left_on_device=$(($(stat -f --format="%a*%S" $out)))
```
nl6720, line 235:
```suggestion
        curr_size="$(stat --format="%s" "$out")"
        space_left_on_device="$(($(stat -f --format="%a*%S" "$out")))"
```

mkinitcpio
```diff
@@ -238,7 +250,13 @@ build_image() {
             sort -z |
             LANG=C bsdtar --uid 0 --gid 0 --null -cnf - -T - |
             LANG=C bsdtar --null -cf - --format=newc @- |
-            $compress "${COMPRESSION_OPTIONS[@]}" > "$out"
+            $compress "${COMPRESSION_OPTIONS[@]}" > "$compressout"
+
+    # sync and rename as we wrote to a tempfile
+    if [[ "$compressout" != "$out" ]]; then
+        sync -d "$compressout"
+        mv -f "$compressout" "$out"
```
nl6720, line 258:
```suggestion
        sync -d -- "$compressout"
        mv -f -- "$compressout" "$out"
```

nl6720 (changes requested):

mkinitcpio
```diff
@@ -228,6 +228,18 @@ build_image() {
...
     esac
 
+    if [[ -f "$out" ]]; then
+        local curr_size space_left_on_device
+
+        curr_size=$(stat --format="%s" $out)
+        space_left_on_device=$(($(stat -f --format="%a*%S" $out)))
+
+        # check if there is enough space on the device to write the image to a tempfile, fallback otherwise
+        if (( $(($curr_size*2)) < $space_left_on_device )); then
```
nl6720, line 238: This probably needs to a little buffer.

nl6720 (changes requested):

nl6720:
I'm getting:
```
==> ERROR: Image generation FAILED: find reported an error
```
The `find` in question is this one:

```bash
    find . -mindepth 1 -printf '%P\0' |
            sort -z |
            LANG=C bsdtar --uid 0 --gid 0 --null -cnf - -T - |
            LANG=C bsdtar --null -cf - --format=newc @- |
            $compress "${COMPRESSION_OPTIONS[@]}" > "$compressout"
```

Which turns into:

```bash
+ find . -mindepth 1 -printf '%P\0'
+ sort -z
+ LANG=C
+ bsdtar --uid 0 --gid 0 --null -cnf - -T -
+ xz --check=crc32 -9e
+ LANG=C
+ bsdtar --null -cf - --format=newc @-
+ [[ /boot/initramfs-linux.img != \/\b\o\o\t\/\i\n\i\t\r\a\m\f\s\-\l\i\n\u\x\.\i\m\g ]]

```

This is with a 100MiB `/boot`:

```
Filesystem     Type  Size  Used Avail Use% Mounted on
/dev/loop0     vfat   99M   81M   18M  83% /boot
```

nl6720:
> ```
> ==> ERROR: Image generation FAILED: find reported an error
> ```

This issue is now gone.

nl6720:
Now it's just a question of how much space to reserve. :thinking:

christian-heusel:
> Now it's just a question of how much space to reserve. thinking

Wdym? Imo we could just leave it as is and maybe emit a warning if the partition is to full 🤔 But since its just not doing a more safe upgrade I think this also could be left as-is.

> The [suggested /boot size in the wiki](https://wiki.archlinux.org/title/Partitioning#/boot) would need updating too.

With the now implemented behaviour (doing atomic rename when possible, otherwise use the current behaviour) I think one could also argue that there is no change needed in the wiki.

mkinitcpio
```diff
@@ -228,6 +228,17 @@ build_image() {
...
     esac
 
+    if [[ -f "$out" ]]; then
+        local curr_size space_left_on_device
+
+        curr_size="$(stat --format="%s" "$out")"
+        space_left_on_device="$(($(stat -f --format="%a*%S" "$out")))"
+
+        # check if there is enough space on the device to write the image to a tempfile, fallback otherwise
+        # this assumes that the new image is not more than 1 1/4x the size of the old one
```
nl6720, line 238:
```suggestion
        # this assumes that the new image is not more than 1¼ times the size of the old one
```

nl6720:
> With the now implemented behaviour (doing atomic rename when possible, otherwise use the current behaviour) I think one could also argue that there is no change needed in the wiki.

:+1:

nl6720:
> Wdym? Imo we could just leave it as is and maybe emit a warning if the partition is to full thinking But since its just not doing a more safe upgrade I think this also could be left as-is.

I mean, is 1¼× enough to account for the new image's potential size increase?

christian-heusel:
> I mean, is 1¼× enough to account for the new image's potential size increase?

🤷🏻‍♂️ I dont know, for my current setup this would mean 8.5MB slack for the fallback img. I think it should be enough for regular upgrades of any kind.

> maybe emit a warning if the partition is to full thinking

what do you think about this?

nl6720 (approved):

nl6720:
It'd feel safer to additionally add some fixed (not dependent on the previous image size) amount of free space to the equation.

christian-heusel:
Re-requesting review because of the bigger changes after the initial approve 😊

Foxboron:
I need to test it so I understand how it works. It looks good to me but please squash the last 3 commits into the 2nd one, or squash everything together :)

christian-heusel:
@Foxboron commits are now squashed 👍🏻

PR #125, opened 2022-10-22
Add --remove flag to presets

### LINKED ISSUE #8: Add a removal flag
Add a -R --remove flag to remove the initramfs images for a given preset. This will make it easier to cleanup after the kernels.

rogueai:
This is still WIP, but I wanted to get the general idea out for review before I get too far. I treated `--remove` as an option to `--preset|--allpresets`.

~~I still have some error handling to manage, e.g: when `rm` fails for some reason (permissions, etc.), I get the usual rm errors, I'd like to add some pretty printed error messages.~~

I'm still debated over the way I changed `process_preset`, I'm open to suggestions!

- [x] add remove flag as an option to `--preset|--allpresets` + error handling
- [x] help/usage, man docs
- [x] test functionality
- [x] test completion

Sample output:
```
❯ ./mkinitcpio -P --remove
==> Removing image for preset: mkinitcpio.d/example.preset: 'default'
==> WARNING: Image not found: `/tmp/arch-linux.efi'
==> ERROR: Image not writable: `/tmp/initramfs-linux.img'
==> Removing image for preset: mkinitcpio.d/example.preset: 'fallback'
  -> Removed: `/tmp/arch-linux-fallback.efi'
  -> Removed: `/tmp/initramfs-linux-fallback.img'
==> Removing image for preset: mkinitcpio.d/hook.preset: 'default'
==> WARNING: Image not found: `/boot/initramfs-%PKGBASE%.img'
==> Removing image for preset: mkinitcpio.d/hook.preset: 'fallback'
==> WARNING: Image not found: `/boot/initramfs-%PKGBASE%-fallback.img'
==> Removing image for preset: mkinitcpio.d/test.preset: 'default'
==> ERROR: Image not writable: `/efi/EFI/arch/arch-linux.efi'
==> ERROR: Image not writable: `/boot/initramfs-linux.img'
==> Removing image for preset: mkinitcpio.d/test.preset: 'fallback'
==> ERROR: Image not writable: `/efi/EFI/arch/arch-linux-fallback.efi'
==> ERROR: Image not writable: `/boot/initramfs-linux-fallback.img'
```

Fixes #8


mkinitcpio
```diff
@@ -443,9 +453,25 @@ process_preset() (
...
+        if (( _optremove )); then
+            if (( ${#preset_remove_cmd[*]} )); then
+                # we won't be calling mkinitcpio recursively, so no need to set MKINITCPIO_PROCESS_PRESET
+                for pr in "${preset_remove_cmd[@]}"; do
+                    if [[ ! -f "${pr}" ]]; then
+                        warning "Image not found: \`%s'" "$pr"
+                    elif [[ ! -w "${pr}" ]]; then
+                        error "Unable to delete image: \`%s'" "$pr"
+                    else
+                        "rm" "-f" "${pr}"
```
nl6720, line 465: Why are `rm` and `-f` quoted?

```suggestion
                        rm -f -- "$pr"
```

mkinitcpio
```diff
@@ -443,9 +453,25 @@ process_preset() (
...
-        msg2 "${preset_cmd[*]}"
-        MKINITCPIO_PROCESS_PRESET=1 "$0" "${preset_cmd[@]}"
+        if (( _optremove )); then
+            if (( ${#preset_remove_cmd[*]} )); then
+                # we won't be calling mkinitcpio recursively, so no need to set MKINITCPIO_PROCESS_PRESET
+                for pr in "${preset_remove_cmd[@]}"; do
+                    if [[ ! -f "${pr}" ]]; then
+                        warning "Image not found: \`%s'" "$pr"
+                    elif [[ ! -w "${pr}" ]]; then
+                        error "Unable to delete image: \`%s'" "$pr"
```
nl6720, line 463: It may be better to explicitly say that it's not writable. "Unable to delete image" just makes one wonder "why?".

mkinitcpio
```diff
@@ -627,6 +656,12 @@ if [[ -t 1 ]] && (( _optcolor )); then
     try_enable_color
 fi
 
+# if we get presets and remove flag, skip to preset processing
+if (( _optremove )) && (( ${#_optpreset[*]} )); then
```
nl6720, line 660:
```suggestion
if (( _optremove && ${#_optpreset[*]} )); then
```

mkinitcpio
rogueai: No reason, fixed

mkinitcpio
rogueai: Updated to:
```
==> ERROR: Image not writable: `/tmp/arch-linux.efi'
```

nl6720 (approved):
Please rebase and squash.

nl6720 (approved):

man/mkinitcpio.8.txt
```diff
@@ -139,6 +139,14 @@ Options for UEFI executable
 	Include a os-release file for the UEFI executable.
 	Default: '/etc/os-release' or '/usr/lib/os-release'.
 
+Options for preset
```
Foxboron, line 142: Remove this section seperation and mention `--remove` inline. Specify it can only be used with `--preset`.

mkinitcpio
```diff
@@ -75,6 +75,9 @@ usage: ${0##*/} [options]
    --kernelimage <path>         Kernel image
    --uefistub <path>            Location of UEFI stub loader
 
+  Options for preset (-p|--preset, -P|--allpresets):
```
Foxboron, line 78: Same as above here.

mkinitcpio
```diff
@@ -443,9 +453,25 @@ process_preset() (
...
+                        rm -f -- "${pr}"
+                        msg2 "Removed: \`%s'" "$pr"
+                    fi
+                done
+            fi
+        else
+            preset_cmd+=($OPTREST)
+            msg2 "${preset_cmd[*]}"
+            MKINITCPIO_PROCESS_PRESET=1 "$0" "${preset_cmd[@]}"
+        fi
```
Foxboron, line 474: Currently `mkinitcpio -P linux -- --uefi "lol.efi"` is going to produce an invocation of `mkinitcpio` with all argument after `--` appended.

This isn't semantically correct and not what people expect should happen.

What I'm wondering is if we should instead record all the flags we are parsing and if we see `-p/-P` we should instead append all of these flags on the preset processing. It would allow us to implement `--remove` as a simple `does the outputfile exist? rm and exit 0`.

I suspect this is larger change, and we should instead accept this for the time being then see if we can rework `mkinitcpio` with the above mentioned change later.

Thoughts @nl6720 ?

Foxboron (changes requested):
Just one change to the manpage and help output :) There is one discussion point but it is somewhat beside this pull request!

mkinitcpio
Foxboron: Then we are on the same page :)

Foxboron (approved):
Please squash the last commit as well :)

PR #126, opened 2022-10-25
Update mkinitcpio.conf

nl6720:
* Use more commonly known modules for the `MODULES` array example.
* Add the `modconf` hook to all example lines.
* Add keyboard, `keymap` and `consolefont` hooks to the examples using encryption. Mostly based on https://wiki.archlinux.org/title/dm-crypt/System_configuration
* Replace `mdadm` hook with `mdadm_udev`. There were merged into one. Fixes https://bugs.archlinux.org/task/68495
* Do not mention the obsolete `shutdown` hook.
* Add `kms` to the default `HOOKS` array.

### DIFF
--- a/mkinitcpio.conf
+++ b/mkinitcpio.conf
@@ -3,7 +3,7 @@
 # The following modules are loaded before any boot hooks are
 # run.  Advanced users may wish to specify all system modules
 # in this array.  For instance:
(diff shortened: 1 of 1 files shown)


Foxboron (approved):
Long term I think it would be better to move these examples into the `mkinitcpio.con(5)` man page.

PR #127, opened 2022-10-25
Missing semicolons

zeroxsevenf:
cause errors with strip hook


nl6720 (approved):
Oops. I missed it in the merge commit.

PR #129, opened 2022-10-25
functions: provide /etc/os-release

nl6720:
Provide `/etc/os-release` and create a symbolic link from `/etc/initrd-release` to it.

Fixes [FS#67454](https://bugs.archlinux.org/task/67454).


functions
```diff
@@ -793,8 +793,20 @@ initialize_buildroot() {
...
+            install -Dm0644 "$osreleasefile" "${buildroot}${osreleasefile}"
+            cp -adT /etc/os-release "${buildroot}/etc/os-release"
+            cp -adT /etc/os-release "${buildroot}/etc/initrd-release"
+        else
+            install -Dm0644 /etc/os-release "${buildroot}/etc/os-release"
+            ln -sT os-release "${buildroot}/etc/initrd-release"
+        fi
+    else
+        >"$buildroot/etc/initrd-release"
+    fi
```
Foxboron, line 809: Hm, I think this "copy or symlink" thing can be usefull other places as well. Maybe something we can make an issue around but not something we should necessaily do in this PR.

Foxboron (approved):

PR #130, opened 2022-10-27
install/kms: add modules that may implement the privacy screen interface

nl6720:
https://lore.kernel.org/intel-gfx/20210421204804.589962-1-hdegoede%40redhat.com/ says that "privacy screen is actually controlled by some vendor specific ACPI/WMI interface which has a driver under drivers/platform/x86".

Since there's also `/drivers/platform/chrome/chromeos_privacy_screen`, check whole `/drivers/platform/` for modules with "acpi", "privacy" or "wmi" in their filename.

Fixes [FS#72964](https://bugs.archlinux.org/task/72964).

### DIFF
--- a/install/kms
+++ b/install/kms
@@ -1,12 +1,16 @@
 #!/usr/bin/env bash
 
 build() {
+    # AGP and DRM modules for GPUs
     map add_checked_modules '/drivers/char/agp/' '/drivers/gpu/drm/'
+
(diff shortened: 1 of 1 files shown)


klausenbusk:
Migrated to to: https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio/-/merge_requests/130.

PR #131, opened 2022-10-28
fsck: Don't add fsck binary for OverlayFS

M3t0r:
When using OverlayFS on the root mount, the fsck hook will attempt to add the corresponding fsck binary, which doesn't exist.

      -> Running build hook: [fsck]
    ==> ERROR: file not found: `fsck.overlay'

My `mount` output on the installation where I encountered this error:

    none on / type overlay (rw,relatime,lowerdir=base/,upperdir=overlay/overlay-upper/,workdir=overlay/overlay-work/)

### DIFF
--- a/install/fsck
+++ b/install/fsck
@@ -5,6 +5,7 @@ build() {
 
     add_fsck() {
         [[ $1 = tmpfs ]] && return
+        [[ $1 = overlay ]] && return
 
         if [[ $1 = ext[234] ]]; then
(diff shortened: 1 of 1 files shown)


nl6720 (approved):

M3t0r:
That was quick 😅

nl6720:
I was thinking about reworking the `fsck` hook yesterday, so it was on my mind.

PR #137, opened 2023-10-02
install/block: include pci controller modules

felixonmars:
Fixes virtio-blk devices attached to PCI on QEMU (which is the case for libvirt), and possibly also hyperv.

Only 38K in size on x86_64, shouldn't be a problem to add :)

### DIFF
--- a/install/block
+++ b/install/block
@@ -25,6 +25,9 @@ build() {
     # spi (mmc in spi mode)
     add_checked_modules '/drivers/spi/'
 
+    # pci controller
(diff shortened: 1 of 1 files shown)


archlinux-github:
This project is a read-only mirror of: https://gitlab.archlinux.org/archlinux/mkinitcpio, please submit your work there.

*This comment was generated by [github-pull-closer](https://gitlab.archlinux.org/archlinux/github-pull-closer).*

PR #139, opened 2025-09-27
Fix check if kernel command line parameter is given as an absolute path

StMartin81:
If I understand the code correctly I think this condition is supposed to check if the value given for the --kernel command line parameter is an absolute path.

### DIFF
--- a/mkinitcpio
+++ b/mkinitcpio
@@ -917,7 +917,7 @@ while :; do
             _optnocmdline=1
             ;;
         -k | --kernel)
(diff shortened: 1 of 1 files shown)


archlinux-github:
This project is a read-only mirror of: https://gitlab.archlinux.org/archlinux/mkinitcpio, please submit your work there.

*This comment was generated by [github-pull-closer](https://gitlab.archlinux.org/archlinux/github-pull-closer).*

PR #140, opened 2026-07-14
functions: support source:destination:mode syntax in FILES=()

berserker-systems:


jelly:
Pull requests should go to https://gitlab.archlinux.org/archlinux/mkinitcpio/mkinitcpio

